Issue #16, opened 2026-07-18
Perf regression: DIDComm exchanges slower since credo/bifold 0.6 upgrade


albertoleon7794:
**Non-blocking** — a reminder to profile later.

Since the credo-ts / bifold 0.6 upgrade, peer-to-peer connections and VRC exchanges are a few seconds slower than the pre-upgrade code.

### Measured (real-device Android, witnessed exchange)
Local processing per DIDComm message — `Agent received message` → decrypted/typed, i.e. *after* network delivery, so it's local cost:

- light (connection/routing) messages: **0.17–0.46 s** each
- heavy (credential / JSON-LD) messages: **1.4–2.8 s** each
- a witnessed exchange moved **~67 messages**; even a plain connection is ~10–15

So local per-message cost alone accounts for the slowdown, before any network.

### Localized (not root-caused)
- **Not the mediator/pickup:** pickup is `Implicit` (persistent websocket); the gap measured is local, after delivery. The self-hosted mediator (aaleon.com) adds round-trips but isn't the dominant cost.
- **Likely credo 0.6 KMS layer:** DIDComm pack/unpack crypto now routes through `KeyManagementModule` → backend dispatch → askar (vs. direct wallet crypto in 0.5). Per-op overhead × dozens of messages/exchange compounds.
- DI signing on-device is separately slow (eddsa-rdfc-2022 / RDFC on Hermes: 841–3654 ms, variable) — new, DI-specific, not the general regression.

### Next steps for whoever picks this up
- Instrument + compare one unpack op on 0.5 vs 0.6 (KMS dispatch + askar key lookup — candidate for caching).
- Separately profile credential JSON-LD / DI processing.
- Check credo upstream for 0.6 DIDComm perf notes.

PR #14, opened 2026-07-06
Upgrade Phases 3+4+5: bifold 3.0.16, credo-ts 0.6.3, React 19, RN 0.81, bc-wallet-mobile ports, VC 2.0

albertoleon7794:
## Summary

Stacked on #13 (Phase 2). Closes the app-layer portion of #10.

### Phase 3 — platform upgrade

- **Deps**: React Native 0.81.5 / React 19.1.0, `@bifold/*` 3.0.16 portals (incl. new `@bifold/react-hooks`), `@credo-ts/*` 0.6.3 with `@credo-ts/didcomm`, askar → `@openwallet-foundation/askar-react-native` 0.6.0, expo modules; Storybook removed
- **Credo 0.6 migration**: `bc-agent-modules` rewritten (`DidCommModule` config, KMS with askar + expo secure-environment), wallet id/key move into the Askar store config, DIDComm type/event renames across hooks/utils, DRPC-based BC attestation flow removed (Keyring doesn't use BC's attestation service)
- **Android**: compileSdk 36, Kotlin 2.1.20, Gradle 8.14.3, **New Architecture enabled**, expo autolinking, lockfiles regenerated
- **iOS**: Podfile expo autolinking, pods on RN 0.81.5
- **Metro/Jest/Babel**: metro 0.83 package-exports config, singleton react resolution for bifold sources, RN 0.81 jest mocks; `react-native-document-picker` patched for RN 0.81; unused `react-native-date-picker` removed
- **Store migration**: Askar 0.2→0.6 store opens in place; legacy R-Card template records backfilled with a placeholder proof (credo 0.6 validation)

### Phase 4 lite — bc-wallet-mobile ports (`dc944fd`)

- **Errors framework** (`app/src/errors/`): `AppError` + trimmed `ErrorRegistry`, handler utilities, `ErrorInfoCard`/`AppErrorModal`/`ErrorBoundary`, `ErrorAlertProvider` wired into `App.tsx`. Snowplow analytics dropped — "Report this problem" sends to Loki with a user-facing reference code
- **`PressableOpacity`** and small hooks (`usePreventGestureBack`, `useDeclineCredentialOffer`, `useDeclineProofRequest`, `useAutoRequestPermission`)
- **Mediator pickup simplified**: unconditional `PickUpV2LiveMode` (WebSocket push); batch pickup, 5s trust-ping polling, and the `MEDIATOR_USE_V2_BATCH_PICKUP` env flag removed (supersedes #7)
[...]

PR #13, opened 2026-07-05
Upgrade Phases 1+2: E2E VRC exchange harness, React Native 0.77.3, drop BCSC

albertoleon7794:
## Summary
- **Phase 1 (#8)**: Appium/WebdriverIO two-wallet E2E harness in `e2e/` — full onboarding plus a complete VRC exchange between an Android emulator and an iOS simulator, using an invitation-URL paste fallback instead of camera QR scanning.
- **Phase 2 (#9)**: React Native 0.73.11 → 0.77.3 (Kotlin 2.0.21, Gradle 8.10.2, AGP 8.7.3, compileSdk 35, iOS target 15.1, Flipper removed), `@react-native-firebase` 14 → 21, `RCTAppDependencyProvider` in AppDelegate, CMake patches for the RN 0.76+ `libreactnative.so` consolidation, and fixes for the safe-area-context 5.x layout regression that hid bottom controls in modal screens.
- Drops the unused BCSC integration (`packages/bcsc-core`, `app/src/bcsc-theme/`, BCSC state/mode) while keeping the keyring theme. KeyRing is now the only mode.
- Bumps the bifold submodule to `e94c45b2` (branch `upgrade/phase2-rn77` on keyring-bifold: E2E invitation hook, RN 0.77 compat, layout fixes).
- credo-ts, bifold version, and React 18 are intentionally untouched — those move in Phase 3 (#10).

Progress log and phase plan: `UPGRADE_PROGRESS.md`.

## Test plan
- [x] `app` jest: 14 suites / 27 tests green
- [x] `bifold` core jest: 134 suites / 1242 tests green (includes VRC conformance tests)
- [x] typecheck + eslint green in both repos
- [x] Android `assembleDebug` and iOS simulator build pass
- [x] Two-device E2E VRC exchange green on Android and iOS

Closes #8. Closes #9.

Made with [Cursor](https://cursor.com)

Issue #1, opened 2026-04-28
Merge in and align with latest BC Wallet


bmiller59:
Merge in the latest code from upstream [BC Wallet](https://github.com/bcgov/bc-wallet-mobile). Establishing this alignment will make contributing code back easier and make our code base more useful to a wider audience.

PR #17, opened 2026-07-18
Major upgrade: RN 0.81 / React 19 / credo-ts 0.6.3, VC 2.0, RCE v3 + eddsa-rdfc-2022 Data Integrity, witnessed exchange

### LINKED ISSUE #3: Bump keyring-bifold for Google attestation root rotation
### Summary

Track the wallet-side integration for Android Key Attestation root certificate rotation and verification hardening in `keyring-bifold`.

Implementation lives in the **bifold submodule** (`packages/react-native-attestation`). This issue covers submodule bump, release notes, and regression validation in the Keyring app.

**Upstream:** https://github.com/berkmancenter/keyring-bifold/issues/20

### Context

Google is rotating attestation roots for Remote Key Provisioning (RKP). The embedded legacy RSA root expires **2026-05-24**. Without updated roots, VRC Secure Exchange hardware attestation verification may fail on newer Android devices and after the legacy root sunset.

Verification path in the app:

```
VRC credential → BiometricSignatureVerifier → native verifyHardwareEvidence
```

No wallet app code changes expected beyond the submodule pointer (unless release versioning / CHANGELOG).

### Acceptance criteria

- [ ] `bifold` submodule bumped to keyring-bifold `main` after PR for issue #20 merges
- [ ] Secure Exchange hardware attestation smoke-tested on Android (RKP if available) and iOS cross-device verify
- [ ] CHANGELOG entry when cutting next release
- [ ] No changes to `.env` or CI secrets required for this work

### Branch

`20-update-google-attestation-roots` (tracks bifold work; submodule bump commits land here)

### Depends on

- [ ] keyring-bifold #20 merged to `main`

### LINKED ISSUE #8: Upgrade Phase 1: Appium two-wallet E2E harness for VRC exchange
Part of the upstream alignment effort (#1). Tracks Phase 1 of the plan in `UPGRADE_PROGRESS.md`.

## Scope
Build an automated end-to-end regression gate before touching any dependencies: two wallets (Android emulator + iOS simulator) run full onboarding and perform a complete VRC exchange against the hosted mediator/witness-server.

## What was built
- `e2e/` WebdriverIO + Appium harness: onboarding flow, connection via invitation URL, VRC issuance + secure exchange between the two devices.
- QR-scan workaround: bifold exposes the invitation URL and a paste-URL fallback on the Scan screen (dev/E2E only), so no camera-to-camera scanning is needed on emulators.
- Runs against debug builds with Metro; apps are reinstalled fresh each run.

## Status
Implemented and green on both platforms (branch `upgrade/phase2-rn77`, commits `4f351d3` root / `1718d331` bifold). This E2E run is the exit gate for every subsequent upgrade phase. Remains open until the branch is merged.

### LINKED ISSUE #9: Upgrade Phase 2: React Native 0.73.11 → 0.77.3, drop BCSC integration
Part of the upstream alignment effort (#1). Tracks Phase 2 of the plan in `UPGRADE_PROGRESS.md`.

## Scope
First dependency hop: bring the app from RN 0.73.11 to 0.77.3 while keeping React 18, credo-ts 0.5.17, and the current bifold untouched. Also removes the BC Services Card (BCSC) integration, which we don't use, while keeping the keyring theme.

## Highlights
- RN 0.77.3 template applied; Kotlin 2.0.21 / Gradle 8.10.2 / AGP 8.7.3 / compileSdk 35; iOS deployment target 15.1, Flipper removed.
- Patched `@hyperledger/{anoncreds,aries-askar,indy-vdr}-react-native` CMake for the RN 0.76+ `libreactnative.so` consolidation.
- `@react-native-firebase` 14 → 21 (v14 incompatible with RN 0.77) + `RCTAppDependencyProvider` in AppDelegate.
- Fixed an RN 0.77 / safe-area-context 5.x layout regression that pushed bottom controls off-screen in several modal screens (bifold + app).
- Deleted `packages/bcsc-core`, `app/src/bcsc-theme/`, BCSC state/mode handling; KeyRing is the only mode.

## Status
Complete on branch `upgrade/phase2-rn77`: jest green in both repos, Android + iOS debug builds pass, and the two-device E2E VRC exchange (#8) is green on both platforms. Remains open until merged.

### LINKED ISSUE #10: Upgrade Phase 3: bifold 3.0.16 + credo-ts 0.6.3 + React 19 + RN 0.81
Part of the upstream alignment effort (#1). Tracks Phase 3 of the plan in `UPGRADE_PROGRESS.md`. Depends on Phase 2 (#9).

## Scope
The big hop — realign the bifold fork with upstream and modernize the stack:

- **bifold**: new branch based on upstream `openwallet-foundation/bifold-wallet` v3.0.16 content; port our custom delta (VRC module, attestation crypto/signing, keyring E2E hooks) onto it. Our fork has no shared git history with upstream, so this is a content re-base, not a git merge.
- **credo-ts 0.5.17 → 0.6.3**: mechanical API rewrite (`agent.credentials` → `agent.modules.credentials`, explicit DIDComm module registration), askar package rename (`@hyperledger/aries-askar-*` → `@openwallet-foundation/askar-*`), re-evaluate existing yarn patches.
- **App**: React 19 + RN 0.81 (reference: `bcgov/bc-wallet-mobile` main); drop Storybook (React 19 incompatible).
- **New Architecture**: attempt `newArchEnabled=true` (upstream default); audit custom native modules (react-native-attestation); fall back to false if needed and record why.
- Wallet-open/migration smoke test (askar store compatibility).

## Exit gate
VRC conformance tests green, jest green in both repos, both platforms build, two-device E2E VRC exchange (#8) green.

### LINKED ISSUE #11: Upgrade Phase 4: app-layer sync with bc-wallet-mobile
Part of the upstream alignment effort (#1). Tracks Phase 4 of the plan in `UPGRADE_PROGRESS.md`. Depends on Phase 3 (#10).

## Scope
With the core stack current, port the app-layer improvements we want from `bcgov/bc-wallet-mobile` main: container/DI structure, screen updates, and any tooling/config changes that make future syncs cheaper. Selective — we take what serves the keyring wallet, not a wholesale copy (BCSC stays out).

## Exit gate
Full test suite green, both platforms build, two-device E2E VRC exchange (#8) green.

### LINKED ISSUE #12: Upgrade Phase 5: move VRC to W3C VC Data Model 2.0
Part of the upstream alignment effort (#1). Tracks Phase 5 of the plan in `UPGRADE_PROGRESS.md`. Depends on Phase 3 (#10).

## Scope
Move the VRC (Verifiable Relationship Credential) module from VC Data Model 1.1 / `Ed25519Signature2018` to VCDM 2.0 / Data Integrity proofs:

- **Prerequisite check**: verify credo-ts 0.6.3 actually supports issuing/verifying JSON-LD Data Integrity VCDM 2.0 credentials (its documented VCDM 2.0 support covers vc+jwt / dc+sd-jwt; the JSON-LD/LDP path is unverified). If not, scope a workaround or defer.
- v2 contexts in `vrc-contexts`, documentLoader updates in `vrc-shared`, witness-server verification updates, RCE protocol version bump.
- **Breaking change, no data migration**: existing VRCs will not be migrated (accepted — near-zero current users).
- DIDComm v2 remains out of scope until stable in credo-ts.

## Exit gate
VRC conformance tests green against VCDM 2.0, witness-server verifies the new proofs, two-device E2E VRC exchange (#8) green.

### LINKED ISSUE #15: VRC: migrate DIDComm JSON-LD proof to DataIntegrityProof (eddsa-rdfc-2022)
## Summary

Follow-up to Phase 5 (#12). We shipped **VCDM 2.0 data model** for VRC/RCard while intentionally keeping **`Ed25519Signature2018`** over Credo DIDComm JSON-LD.

This issue tracks a **separate, deeper** migration to W3C Data Integrity:

- `type: DataIntegrityProof`
- `cryptosuite: eddsa-rdfc-2022` (or current W3C-aligned suite)
- Do **not** migrate 2018 → `Ed25519Signature2020`

## Why deferred

Documented in [`docs/CRYPTO_SUITE_FOLLOWUP.md`](https://github.com/berkmancenter/keyring-wallet/blob/upgrade/phase3-bifold3/docs/CRYPTO_SUITE_FOLLOWUP.md):

1. Credo DIDComm `jsonld` path is Linked Data Signatures / 2018 today.
2. Credo 0.6 `W3cV2CredentialService` is JWT / SD-JWT oriented — not DI-over-DIDComm for our exchange.
3. Interop with existing wallets, stored RCards, and witnesses.
4. Needs a **clean upgrade state** before a profound crypto-path change.

## Suggested scope (when ready)

- [ ] Spike: can Credo 0.6 (or newer) verify/issue DI proofs on the DIDComm credential exchange path we use?
- [ ] Dual-stack or migration period for stored 2018 credentials
- [ ] Witness / peer interop plan
- [ ] Update DTG guidance to prefer DI; keep 2018 as legacy-acceptable
- [ ] E2E (sim + real device) for Secure Exchange under DI

## Non-goals

- Hardware attestation evidence (P-256 SE/TEE) — orthogonal; see `docs/HARDWARE_ATTESTATION_FLOW.md`
- Spec example “Signature2020” as an intermediate suite

## Related

- #12 (Phase 5 — data model)
- bifold: comment on `proofType: 'Ed25519Signature2018'` in `packages/core/src/modules/vrc/vrc-manager.ts`

albertoleon7794:
> **📖 Reviewing this PR?** Start with the runbook: [`docs/REVIEWING_THE_UPGRADE.md`](docs/REVIEWING_THE_UPGRADE.md) — setup, static gates, the E2E ladder (and what each run proves), security invariants, and a suggested review order.

Consolidates the full upgrade effort into one branch for review. Supersedes
the stacked upgrade PRs #13 and #14 (closed in favor of this).

Scope (29 commits; see each ticket for design/details):
- Phase 1 — Appium two-wallet E2E harness. Closes #8
- Phase 2 — RN 0.73→0.77, drop BCSC. Closes #9
- Phase 3 — bifold 3.0.16, credo-ts 0.6.3, React 19, RN 0.81. Closes #10
- Phase 4 — app-layer sync with bc-wallet-mobile. Closes #11
- Phase 5 — W3C VC Data Model 2.0 for VRC/RCard. Closes #12
- RCE v3 + eddsa-rdfc-2022 Data Integrity issuance (capability-gated,
  dual-verify with 2018), witnessed-exchange DI + VWC. Closes #15
- Google/Apple attestation root rotation. Closes #3

Proven on real devices: VC 2.0 credential, eddsa-rdfc-2022 DI-signed,
hardware-attested, and witnessed (both contact shields).

Companion bifold PR: berkmancenter/keyring-bifold#31

Known follow-up (non-blocking): DIDComm perf regression since 0.6 — #16.

Reviewer note: the app diff is the review target. The bifold submodule is a
fork-swap (old fork -> bifold-3.x), ~2,257 commits of fork history, not new code.


bmiller59:
## DRY follow-ups in the VRC module (non-blocking)

While reviewing the VRC exchange flow (`bifold/packages/core/src/modules/vrc/**`), a few duplications are worth cleaning up in a follow-up — none affect correctness, they're maintainability/DRY notes:

1. **`counterpartySpeaksVc20` / `counterpartySpeaksDi` in `vrc-manager.ts`** are near-identical: same `RelationshipDidRepository` lookup, same `?? 1` default, differing only in the version threshold (`>= 2` vs `>= 3`) and both swallowing errors to `false`. Suggest collapsing into a single `counterpartyRceVersionAtLeast(agent, counterpartyRelationshipDid, minVersion: number)` helper.

2. **`buildVrcCredential` (`vrc-manager.ts`) and `buildRCardCredential` (`services/rCardCredential.ts`)** duplicate the same `useVc20`/`useDi` → `@context` array selection logic (VCDM2+DI vs VCDM2+Ed25519-suite-context vs legacy VCDM1.1), each with its own copy of the "why" comment. A shared `selectCredentialContexts({ useVc20, useDi }, ...)` helper would remove the duplication and the risk of the two credential builders drifting on proof-context rules as the DI/VC2.0 rollout continues.

3. **`credential.encoded` raw-JSON extraction** (getting the credential JSON out of a `W3cCredentialRecord`, tolerating both plain objects and class instances with `.toJSON()`) is implemented independently four times: `utils/rcardDisplayUtils.ts` (`toRawCredential` — the most careful version, added in this PR), `utils/witnessCredentialUtils.ts` (six inline copies, no `.toJSON()` handling), `utils/vrcNameHelper.ts` (two more inline copies), and `hooks/useDeleteContact.ts` (one more). Since the robust `toRawCredential` helper is new code in this very PR, it'd be worth exporting it and pointing the other three files at it rather than keeping their own inline `credential.encoded` reads.

Happy to file these as a follow-up issue if that's preferred over addressing inline.

bmiller59:
## `vrc-reference` doesn't exercise the app's new capability-negotiation surface

Comparing `bifold/packages/core/src/modules/vrc/**` (the app) against `bifold/packages/vrc-reference/**` (the reference/conformance package) across this PR (`main`@`74ac281d` → this branch@`1755fbb`):

**Aligned:** the credo-ts 0.6 migration (`agent.connections` → `agent.modules.didcomm.connections`, `CredentialEventTypes` → `DidCommCredentialEventTypes`, etc.), relationship-DID creation (`PeerDidNumAlgo.InceptionKeyWithoutDoc` + the same credo-0.6 KMS key-lookup re-import fix), the VCDM 2.0 shape swap (`issuanceDate`→`validFrom` + v2 context), and the witnessed-flow proof-family mirroring (`Witness.getMirroredJsonLdProofOptions` mirrors whatever proof type it's shown, same design as the app's `getVrcJsonLdProofOptions`) are all consistently applied in both packages. These parts look like a safe, faithful upgrade relative to `vrc-reference`'s pre-PR behavior — no leftover old-API calls, no shape mismatches found.

**Gap — the app's new headline feature isn't represented in `vrc-reference` at all:**

1. **No RCE version negotiation.** The app announces/stores a capability version (`vrc:rceVersion:<n>`) per counterparty and gates VC2.0/DataIntegrity issuance on it (`counterpartySpeaksVc20`/`counterpartySpeaksDi` in `vrc-manager.ts`). `Participant.ts` has no equivalent — it unconditionally builds VCDM 2.0 and unconditionally signs with `'Ed25519Signature2018'` at all three call sites. It never issues a v1-legacy credential and never issues a DataIntegrityProof credential from the peer-to-peer issuer path.
2. **No RCard.** `RelationshipCard`/`buildRCardCredential` — the contact-info credential the app now exchanges separately from the VRC — doesn't exist anywhere under `packages/vrc-reference` (zero grep hits outside the app module).
3. **Handshake format diverged further.** The app's basic-message handshake gained an `rceVersion` field this PR; `Participant.ts`'s `JSON.stringify({ rDid: did })` handshake has no field to carry it, so `vrc-reference` can't even represent the new capability announcement, let alone negotiate on it.
4. **DI issuance is only proven in isolation.** The new `diConformance.test.ts` is a single in-memory-agent crypto round-trip (sign → verify → tamper-check), not an end-to-end DIDComm exchange. Because `Participant.ts` never sends a DI-signed VRC, the Witness's `vrcUsesDi` branch is unreachable from any of the real integration tests (`connectionFlow`, `witnessedFlow`, `credentialIssuance`, etc.) — only from that synthetic single-agent test.

**Why this matters:** `vrc-reference` tests passing is good evidence the credo 0.6 migration and the DI cryptosuite adapter are correct in isolation, but it is not evidence that the app's actual new feature — RCE v1/v2/v3 capability negotiation driving dual-stack proof selection over a real DIDComm exchange, plus the RCard split — works end-to-end, since `vrc-reference` doesn't implement or test that surface. Given `vrc-reference` is likely to become the foundation of a conformance test suite, this seems worth closing before relying on it for that purpose.

**Suggested follow-up:** update `vrc-reference`'s `Participant` to negotiate `rceVersion` and issue DI-proof VRCs the same way the app does, and add an RCard-equivalent exchange — bringing it to full parity with the app's negotiated capability surface, while keeping the already-confirmed-safe parts (0.6 migration, VC2.0 shape, proof mirroring) as-is. If `vrc-reference` is meant to deliberately stay v1/v2-only for now, that scope limitation should be stated explicitly in `BIFOLD_FLOW_ALIGNMENT.md` so reviewers don't read "vrc-reference tests pass" as end-to-end conformance proof for the RCE v3 / RCard rollout.

bmiller59:
## Credential display architecture: the RCard split needs a cross-credential composition path

Looking at how VRC/RCard content is displayed (`ContactDetails.tsx`, `ListContacts.tsx`, `CredentialOffer.tsx`, `ListCredentials.tsx`), the original intent — that a credential type's display data is assembled into one object regardless of which underlying credential(s) it comes from — has only partially survived the RCard split, and the gap has already produced a concrete inconsistency worth fixing.

### What exists today

Two separate, non-communicating systems answer "what should this contact/credential look like":

1. **`CredentialDisplayHandler` registry** (`display/displayRegistry.ts`, `RelationshipCredentialHandler`, `WitnessCredentialHandler`) — a clean per-type strategy pattern, but `canHandle`/`extractFields` operate on exactly **one credential JSON** at a time. There's no way for a handler to look at sibling credentials.
2. **`resolveContactDisplayInfo`** (`utils/rcardDisplayUtils.ts`) — a hand-rolled function that scans *all* `w3cCredentialRecords`, re-implements its own DTGCredential/RelationshipCredential/RCard type-matching (duplicating `isDTGCredential`/`isRelationshipCredential` from the registry file), and merges `name`/`email`/`organization` from whichever credential (RCard or legacy embedded issuer object) has them. It's called ad hoc from four separate screens (`ContactDetails.tsx`, `ListContacts.tsx`, `CredentialOffer.tsx`, `chat-messages.tsx`).

The RCard split introduced a genuinely new requirement — one contact's display identity now legitimately spans two W3C credentials (the VRC + the RCard) — but instead of extending the registry to handle that, a parallel resolver was bolted on beside it. Concretely, this already causes drift: in `CredentialOffer.tsx`, the header name is resolved via `resolveContactDisplayInfo` (RCard-aware) while the field list below it is resolved via `RelationshipCredentialHandler.extractFields` (not RCard-aware — reads the embedded issuer object only). For a new-style (v2/v3) offer, the header shows the right name while the field list's "Issuer Name/Email/Organization" rows simply don't render. Not a data-correctness bug, but a symptom of the same information being computed two different ways in the same screen.

### Recommended fix

Extend the existing handler architecture rather than working around it, so there's one doorway instead of two:

**1. Add cross-credential context and a `subject` output to the handler interface** (`display/types.ts`):
```ts
export interface CredentialDisplayContext {
  /** Sibling W3C records in the wallet, for handlers that need to cross-reference
   *  (e.g. RelationshipCredentialHandler resolving a contact's RCard). */
  relatedRecords: W3cCredentialRecord[]
}

export interface CredentialDisplaySubject {
  id: string
  name?: string
  email?: string
  organization?: string
}

export interface CredentialDisplayHandler {
  extractFields(credential: W3cCredentialJson, context: CredentialDisplayContext): Field[]
  /** Optional: canonical display identity for this credential's issuer/subject —
   *  used for headers, chat messages, contact screens. Handlers with no notion
   *  of "subject" (e.g. WitnessCredentialHandler) simply omit it. */
  extractSubject?(credential: W3cCredentialJson, context: CredentialDisplayContext): CredentialDisplaySubject
  // ...unchanged
}
```

**2. Thread it through `CredentialDisplayResult` / `getDisplayInfo`** (`display/displayRegistry.ts`):
```ts
export interface CredentialDisplayResult {
  fields: Field[]
  buttonText: CredentialButtonText
  matched: boolean
  credentialTypeName?: string
  subject?: CredentialDisplaySubject   // NEW
}

getDisplayInfo(credential: W3cCredentialJson, context: CredentialDisplayContext): CredentialDisplayResult {
  const handler = this.findHandler(credential)
  if (!handler) return { fields: [], buttonText: DEFAULT_BUTTON_TEXT, matched: false }
  return {
    fields: handler.extractFields(credential, context),
    subject: handler.extractSubject?.(credential, context),
    buttonText: handler.getButtonText(),
    matched: true,
    credentialTypeName: handler.getCredentialTypeName?.(),
  }
}
```

**3. Move (not delete) the RCard-resolution logic into `RelationshipCredentialHandler`:**
```ts
extractSubject(credential: W3cCredentialJson, { relatedRecords }: CredentialDisplayContext): CredentialDisplaySubject {
  const issuer = extractIssuerObject(credential)
  const rcardInfo = resolveContactDisplayInfo(relatedRecords, issuer.id) // existing, tested logic — just relocated
  return {
    id: issuer.id,
    name: rcardInfo.name ?? issuer.name,
    email: rcardInfo.email ?? issuer.email,
    organization: rcardInfo.organization ?? issuer.organization,
  }
}
```
`rcardDisplayUtils.ts`'s jCard parsing and its existing test suite stay as-is — only the entry point moves.

**4. Update the four call sites** (`CredentialOffer.tsx` header, `ContactDetails.tsx`, `ListContacts.tsx`, `chat-messages.tsx`) to read `.subject` off the one `getDisplayInfo()` call instead of separately calling `resolveContactDisplayInfo`. Header and field list become structurally guaranteed to agree, since they're now sourced from the same function call.

### Why this shape

- **Additive, not a breaking rewrite** — `extractSubject`/`context` are optional/new parameters; handlers that don't need cross-referencing are unaffected.
- **Preserves the registry's real strength** (the core/VRC decoupling via `ICredentialDisplayRegistry`) rather than introducing a second architecture alongside it.
- **Closes the existing duplication** — `isDTGCredential`/`isRelationshipCredential` vs. the type-matching re-implemented in `resolveContactDisplayInfo` collapse into one call path.
- **Regression-proofs itself** — a test asserting `getDisplayInfo(vrcCredential, { relatedRecords }).subject.name` matches what a screen renders as its header would fail immediately if a future change reintroduces a second, drifting path.

Scope: `display/types.ts`, `display/displayRegistry.ts`, `RelationshipCredentialHandler.ts`, and the four call sites above, plus tests. A moderate, well-scoped follow-up — not something that should block this PR, but worth filing before more fields or another related-credential-type need the same cross-referencing.

bmiller59:
## `witness-server`: dual-stack design is sound, but the test suite doesn't exercise the changed behavior

Compared `bifold/packages/witness-server/**` between `main`@`74ac281d` and this branch@`1755fbb` (7 commits, `WitnessService.ts` is the bulk of it at 462 changed lines).

### The good news: backward compatibility holds

Dual-verify for `Ed25519Signature2018` vs `DataIntegrityProof`/`eddsa-rdfc-2022` was implemented by *registering* a new signature suite into Credo's existing suite-resolution machinery (`DataIntegritySuiteModule`), not by rewriting verification logic — so the verification path for old-style VRCs is untouched. Issuance mirrors the same way: `getMirroredJsonLdProofOptions` explicitly defaults to the old hardcoded `Ed25519Signature2018`/VC1.1 options for anything that isn't a recognized DI proof, so an old client's VRC still produces a byte-identical old-shape VWC (`WitnessService.ts` VC2.0 branch is gated on `vrcContexts[0] === CREDENTIALS_V2_CONTEXT_URL`; legacy VRCs fall through to the pre-existing `@context`/`issuanceDate`/`expirationDate` shape unchanged). The freshness-window widening (5→10 min) is a coordinated fix matching a backdating allowance already present on the app side (`vrc-manager.ts`), not an arbitrary loosening. This is genuine dual-stack support, not a bolt-on.

### Finding 1 (moderate-high): the test suite doesn't test the changed code

`WitnessService.test.ts` (2524 lines) wasn't touched in this PR despite the 462-line source diff. On inspection, it's not just stale — it's structurally decoupled from the real implementation:

- The "Freshness Check" tests hardcode the *old* 5-minute window and assert a 6-minute-old credential is rejected. That assertion is now **false** against the shipped code (which accepts up to 10 minutes) — but the test still passes because it only checks its own local arithmetic, never `handlePresentationSubmission`'s actual freshness logic.
- The "VWC Building" tests call a local `buildMockVWC()` helper that still canonicalizes with the *old* `JSON.stringify(vrcJson, Object.keys(vrcJson).sort())` (pre-`jcsCanonicalize`) and builds `issuer` as a bare string, while the real `buildWitnessCredential` now emits `issuer: { id, name }`. Neither a digest-algorithm regression nor an issuer-shape regression would be caught by this test.
- Nothing in the file references `getMirroredJsonLdProofOptions`, `CREDENTIALS_V2_CONTEXT_URL`, `vrcIsVc20`, or DI at all.

The mirroring helper itself is well-covered elsewhere (`vrc-reference/__tests__/unit/mirroredProofOptions.test.ts`, imported directly from `@bifold/vrc-shared`), and DI sign/verify round-trips are covered in isolation by `diConformance.test.ts` — but neither touches `WitnessService.ts`'s two call sites (offer-side proof options, VWC-build data-model branch). Net: a real behavioral change to the core witness flow shipped with a test file that looks like coverage (parallel "VWC Building"/"Freshness Check" describe blocks) but wouldn't fail if this exact diff were wrong.

**Suggestion:** either add tests that call `WitnessService`'s real methods directly for the mirroring/freshness/digest changes, or at minimum fix the now-incorrect 5-minute freshness assertion so it doesn't quietly assert stale behavior.

### Finding 2 (minor): stale, self-contradicting comment

`WitnessService.ts` (added in the DI-verify commit) still has a comment reading *"Witness ISSUANCE stays 2018 until the peer matrix is proven"* — directly contradicted two commits later in the same PR when DI issuance via mirroring was added. Low functional risk (it's a comment), but likely to mislead the next reader into assuming issuance never emits a DI proof.

Everything else checked out clean: `TlsManager`/`LocalityService` diffs are mechanical type/lint fixes with no TLS-behavior change, `index.ts`'s askar-nodejs import-order fix is a well-explained narrow correctness fix, and `CredentialRegistry.ts` (unchanged) doesn't need anything the new mirroring code doesn't already provide.

Non-blocking overall — the design itself is sound — but the test-coverage gap in particular seems worth closing before relying on this suite as a regression guard for the witnessed-exchange flow going forward.

bmiller59:
## `app/src/services/attestation.ts`: unrelated to hardware attestation — but the DRPC removal leaves a silent dead-end

Worth clarifying scope first: despite the name and directory, this file has **nothing to do with hardware/biometric attestation** (Secure Enclave, StrongBox, App Attest, `useHardwareAttestation`) — all of that lives in `bifold/packages/core/src/modules/vrc/*` and has already been covered in earlier review passes. This file is a separate, older subsystem inherited from the upstream BC Wallet fork lineage: `AttestationMonitor` auto-detects and auto-handles proof requests asking for a **remote attestation credential** from BC Government's Traction service, gated by hard-coded `credDefIDs`/invitation URLs per environment in `app/src/constants.ts:14-30` (pointing at `traction-acapy-*.apps.silver.devops.gov.bc.ca`). It's live-wired, not dead code: registered in the DI container (`app/container-imp.ts:111`), started/stopped on every agent lifecycle transition (`app/src/hooks/useBCAgentSetup.ts:63-69`), and subscribed to by `Connection.tsx`/`ProofRequest.tsx` for loading/toast UI.

### Finding (medium): notification filter and failure path are now out of sync

Diffed against `main`: this PR deliberately strips the entire remote-attestation fetch/generate flow (`connectToAttestationAgent`, `fetchNonceForAttestation`, `generateAttestation`, `requestAttestation`, plus the associated `@bifold/react-native-attestation` and `react-native-device-info` imports) — a reasonable simplification, with a comment explaining there's no stable `@credo-ts/drpc` for Credo 0.6 and that Keyring doesn't run the BC service. `requestAttestationCredential` (`attestation.ts:188-204`) now unconditionally logs a warning and fails.

The problem: `useNotifications` (`app/src/hooks/notifications.ts:110-120`) still unconditionally filters proof requests matching `isProofRequestingAttestation` **out of the notification list**, regardless of whether they can be resolved. Combined with the now-guaranteed failure, **any live proof request matching those credDefIDs becomes invisible to the user and permanently unresolvable** — not shown as a notification, not auto-completed. The user only learns something went wrong if they happen to have `Connection.tsx` or `ProofRequest.tsx` mounted at the exact moment the `FailedRequestCredential` event fires; otherwise it's a silent dead end.

Whether this is actually reachable depends on whether Keyring still talks to real BC Government/Traction verifiers gated on these specific credDefIds — the repo still carries BC "Person credential" flow code (`keyring-theme/features/person-flow`, `BCIDHelper.ts`, `CRED_HELP_ACTION_OVERRIDES` in `app/container-imp.ts`), so this isn't provably dead, just unclear.

**Suggested resolution — pick one:**
1. If Keyring still needs this flow, restore a real handler (or an explicit "not supported, tell the user" UI) rather than a silent failure.
2. If Keyring no longer talks to BC's Traction instance at all, remove `AttestationMonitor`/`AttestationRestrictions` and the notification-hiding filter wholesale, rather than leaving a stubbed, dead-ended version wired live in the container.

### Everything else checked out clean

The Credo 0.6/`@credo-ts/didcomm` migration in this file itself is mechanically solid: `agent.proofs.*` → `agent.modules.didcomm.proofs.*`, `proofRecordId`/`credentialRecordId` → `proofExchangeRecordId`/`credentialExchangeRecordId`, `ProofState`/`CredentialState` → `DidCommProofState`/`DidCommCredentialState`, event payload key rename — all consistent with the same pattern used elsewhere (`app/src/utils/credentials.ts`). Lint is clean, no leftover unused imports from the stripped code path. Two low-priority items not worth a blocking discussion: six of nine `AttestationErrorCodes` are now unused dead code left over from the removed flow, and a pre-existing (not introduced by this PR) theoretical stuck-spinner edge case if `handleProofRequest` ever returns `false` with an undefined agent — practically unreachable today, just worth a follow-up ticket if anyone's doing cleanup in this area.

Non-blocking, but the notification/failure mismatch in particular seems worth a deliberate decision rather than leaving as-is.

bmiller59:
## E2E harness review: solid and trustworthy, with two documentation-completeness gaps

The entire `e2e/` package is brand-new on this branch (`git diff --stat main...HEAD -- e2e/` is 100% insertions across every file), so this isn't a "did it preserve old behavior" review — there's no previous harness to compare against. Instead: is this newly-built ~2,350-line Appium harness reliable, and does it actually prove what `docs/REVIEWING_THE_UPGRADE.md` §4 claims each script proves.

### Overall verdict: a green run is trustworthy

No tautological assertions, no try/catch blocks that swallow real failures into a false "pass," and no retry logic that silently masks failures past a real check — every retry loop is bounded and throws a descriptive error if the underlying condition isn't met. Cleanup is consistently correct: sessions, Appium, metro, and the witness server are all torn down in `finally` blocks on both success and failure paths, so a failing run doesn't leak processes. The 5 top-level scripts share real logic through `lib/flows.js`/`lib/driver.js` rather than copy-pasting; the duplication that does exist (device-discovery/preflight helpers between the two `:devices` scripts) is explicit and intentional, not drift.

Script-by-script, checked against the runbook's claims:
- `run-onboarding-smoke.js` — proves exactly what it claims, nothing more.
- `run-store-migration.js` — **non-trivial and correct**: installs a genuinely separate old build (credo 0.5.17/askar 0.2.3), runs a real bidirectional exchange so the old-format store holds real credential data, confirms it *before* upgrading, then does an in-place `adb install -r` (not wipe-and-reinstall) with the new build and re-asserts survival. This is the real "old store opened by new code" test, not a same-version create-and-read.
- `run-vrc-exchange.js` — correctly omits attestation calls; doesn't overclaim what the emulator/simulator can prove.
- `run-vrc-exchange-devices.js` — `expectAttestation: true` actually throws if the `AttestationVerified` banner doesn't appear (a real, failable gate), and `acceptSystemAlertIfPresent` (`driver.js:139-159`) explicitly refuses to touch any biometric/passcode/Face-ID/PIN alert, correctly leaving those for the human operator instead of auto-dismissing them.
- `run-vrc-exchange-witnessed-devices.js` — requires both wallets connect to a real witness (confirmed via the witness's own participant-count log), requires attestation, and `assertContactShields` requires **both** shields present simultaneously before passing. See finding #1 for the one gap here.

### Finding 1 (medium): witness locality-verification is silently disabled by default in the witnessed e2e, and the runbook doesn't say so

`e2e/lib/witness.js:143` — `WITNESS_LOCALITY_REQUIRED` defaults to `"false"`, disabling the witness's BLE co-location check for every witnessed e2e run. This is a deliberate, commented tradeoff (Appium-driven phones can't produce a real locality proof), but `docs/REVIEWING_THE_UPGRADE.md` §4 only documents the attestation-emulator caveat — not this one. A reviewer reading only the runbook would reasonably believe a green `yarn e2e:vrc:witnessed:devices` run proves the full witness protocol, when the locality-verification leg is opted out by default. Worth adding a line to the runbook's E2E ladder table alongside the existing attestation caveat.

### Finding 2 (low): dangling doc reference

`e2e/lib/flows.js:499`, `e2e/lib/witness.js:5`, and `e2e/run-vrc-exchange-witnessed-devices.js:15` all cite `docs/spikes/witnessed-e2e-spec.md` as the design source of truth for the witnessed-exchange harness — that file doesn't exist anywhere in the repo (checked both `keyring-wallet` and the `bifold` submodule). Design rationale for some non-obvious choices (e.g. the locality bypass above) currently lives only in code comments, with no citable doc backing it.

### Finding 3 (low): `debug-witness-connect.js` is scratch code, not a maintained part of the harness

It self-identifies as a "Throwaway debug script" in its own header comment, isn't wired into `e2e/package.json` or the root `package.json` scripts, and isn't documented in `e2e/README.md` — unlike the other 5 scripts. It does serve a real, narrow diagnostic purpose (isolating the witness-connect step without needing a second device or biometric hardware), so it's worth keeping in some form rather than just deleting — but as committed it's an undocumented capability nobody but its author would know to reach for. Suggest either moving it under something like `e2e/scripts/debug/` with a one-line README pointer, or removing it.

### A few low/nit items, not blocking

- `driver.js:71-73` — `stopAppium()` sends `SIGTERM` without awaiting exit or closing the `artifacts/appium.log` write stream. Harmless in practice (process exit flushes it, Appium reliably dies from the signal), but worth a follow-up if Appium ever starts ignoring SIGTERM.
- `witness.js:100-110` — frees leftover processes on ports 9002/9003 via `lsof` + `kill -9` scoped only by port, not pid ownership/cmdline. Fine for a local dev-only harness on fixed test ports.
- `flows.js:558-585`/`:592-612` — shield assertions fall back to a substring text match (`byTextContains`) when preferred testIDs aren't present; low risk since it's a secondary check behind an already-asserted state transition, but looser than the testID-based checks elsewhere.
- `run-vrc-exchange.js` never prints a runtime note that attestation wasn't exercised on that particular run — someone reading only console output (not the docs) on a green run could miss the caveat.

Non-blocking overall — this is a well-built harness. The locality-verification caveat (#1) is the one item I'd actually want addressed before treating a green witnessed-devices run as full proof of the witness protocol.

bmiller59:
## Two things worth calling out explicitly — solid work, not just "no issues found"

Most of the review passes on this PR turned up findings worth discussing (posted as separate comments). These two didn't, and they're worth recognizing on their own rather than going unmentioned.

### VRC exchange core correctness: backward compatibility genuinely held up

A full pass over the VRC exchange core (`vrc-manager.ts`, `witnessed-vrc-manager.ts`, hardware signing/biometric, `EvidenceBuilder`, document loaders, RCard issuance, name resolution, repositories) came back clean. The dual-stack backward-compatibility work — RCE v1/v2/v3 gating, legacy issuer-object fallback for pre-embedded-name credentials, biometric→passcode fallback, legacy evidence-shape validation — is consistently and correctly preserved throughout, not just handled in the obvious spots. For a PR that bundles a framework upgrade with three credential features into one unit, that's a real achievement — it would have been easy for backward-compat handling to be inconsistent across call sites, and it isn't.

### The Google attestation trust-anchor change is a genuine security fix, done right

`bifold` commit `88f80fe8` ("align Google attestation trust anchors with published roots") looks on its face like routine root-rotation maintenance ahead of the legacy RSA root's 2026-05-24 expiry, but it's more than that: the previous code also added the *presented certificate chain's own terminal certificate* as a trust anchor (`certs.last()`), which meant a chain terminating in an attacker-controlled self-signed certificate could pass PKIX validation regardless of whether it actually chained to a real Google root — a real attestation-bypass vulnerability, not just an expiring-root maintenance issue. This PR fixes it properly: `GoogleAttestationChainValidator`/`GoogleAttestationRoots` now validate strictly against the three published Google roots (legacy RSA, re-signed RSA, RKP ECDSA) with byte-exact comparison, with the terminal-cert exclusion applying only when it's independently verified to already be a recognized root. It's mirrored correctly on both Android (Kotlin) and iOS (Obj-C++), and there's a regression test that names the exact bug it closes:
```kotlin
fun validateChain_selfSignedTerminalRoot_failsWithoutSelfAsAnchor() {
  // Regression: trusting certs.last() as an anchor let self-signed terminal roots pass.
  ...
}
```
That's the right way to land a security fix — not just closing the hole, but leaving a test that documents why it mattered so it can't silently regress.

bmiller59:
## Credential versioning/fallback: the negotiation core is solid; the type-detection layer needs consolidation before a v4 shows up

Zooming out across everything reviewed on this PR so far, here's an assessment of whether the old-version/new-version fallback handling is maintainable going forward — not just correct today.

### What's good, and worth keeping as the pattern

- **`rceVersion` as a monotonically increasing integer**, announced on the handshake and stored per-counterparty (`RelationshipDidRecord.counterpartyRceVersion`), is the right shape for this. Adding a v4 doesn't require a wire-protocol change — just a higher number and a new threshold check.
- **Proof-type selection is genuinely centralized.** `getVrcJsonLdProofOptions`/`getMirroredJsonLdProofOptions` (`@bifold/vrc-shared`) is the single function both `vrc-manager.ts` and `witness-server`'s `WitnessService.ts` call to decide `Ed25519Signature2018` vs `DataIntegrityProof`. Dual-verify was achieved by registering a suite into this one shared resolution path rather than forking verification logic — that's exactly the pattern you want for the next crypto-suite addition.

### What isn't contained enough: credential-type detection is reimplemented everywhere

Checking how many places independently string-match "is this a DTGCredential / RelationshipCredential / RCard / RCardTemplate":

```
ListCredentials.tsx        — 3 separate inline copies within the same file (lines 111, 168, 188)
CredentialOffer.tsx        — 2 inline copies
chat-messages.tsx          — 2 inline copies
ListContacts.tsx           — 1 inline copy
rcardDisplayUtils.ts       — 2 inline copies
vrc-manager.ts             — 2 inline copies
witnessCredentialUtils.ts  — 2 inline copies
vrcNameHelper.ts           — 1 inline copy
displayRegistry.ts         — 2 exported helpers (isDTGCredential/isRelationshipCredential)
credential-display.ts      — 1 more, near-duplicate of displayRegistry's (isDTGCredentialType)
```

That's 12+ independent reimplementations across 9 files of the same type-string matching — and notably, none of the ad hoc call sites use the two canonical helpers that already exist and are exported (`displayRegistry.ts`'s `isDTGCredential`/`isRelationshipCredential`). This is the same shape of problem already flagged at two other layers in this PR: the `counterpartySpeaksVc20`/`counterpartySpeaksDi` duplication (DRY comment) and the registry-vs-`resolveContactDisplayInfo` split (credential display architecture comment) — three independent instances of "the same version/shape distinction gets re-derived per call site instead of computed once."

This isn't a current-correctness problem (nothing found where it's actually wrong today), but it's a maintainability one: adding a v4 credential type or a new shape distinction means touching 9+ call sites with nothing — no shared constant, no exhaustive test, no type-level enum — forcing anyone to remember all of them. It's exactly the pattern that already caused the `CredentialOffer.tsx` header/field-list drift found earlier in this review: two independent paths deciding "is this available" differently.

### Suggested follow-up issue (non-blocking on this PR)

Two scoped, mechanical fixes:

1. **Consolidate the 12+ type-check reimplementations into one canonical module** (extend `displayRegistry.ts`'s existing exports, or a new `credentialTypes.ts`) — `isDTGCredential`, `isRelationshipCredential`, `isRCard`, `isRCardTemplate`, `isWitnessCredential` — and update all call sites to import from it instead of inlining string-literal arrays.
2. **Add a version-compatibility matrix test** — one test suite that runs the credential-building/display/witness-verification pipeline across the cross product of `{v1, v2, v3} × {DI, Ed25519} × {witnessed, non-witnessed}` and asserts each combination produces the expected shape. Nothing today would catch a new version silently breaking an old path — this would also have caught the witness-server freshness-test drift flagged in an earlier comment.

Filing this as a heads-up for planning rather than something to resolve in this PR — the negotiation primitive and proof-mirroring centralization are the hard parts and they're already done right; this is cleanup that gets more expensive the longer it waits.

bmiller59:
## Dependency patches: sound and minimal, but a few maintainability items worth tracking going forward

Reviewed all 9 `.yarn/patches/` entries plus `app/patches/rdf-canonize+5.0.0.patch`, reading each patch's actual diff content and cross-checking several against real upstream repo/npm state.

### The patch set itself is solid

Every patch is small, single-purpose, and has a verifiable, non-hacky justification:

- **`react-native-npm-0.81.5`** (TextInput memory-leak fix) — verified as a real, merged upstream commit (`facebook/react-native#55719`) that hasn't shipped in any released RN version yet (merged after the last 0.81.x tag; still not in 0.82.x either). Not "a patch instead of a version bump" — no version to bump to exists yet.
- **`@hyperledger/indy-vdr-react-native`** (int sentinel `0`→`-1`) — a real bug, not cosmetic: `0` is a legitimate value for parameters like pool-submit timeout, so using it as a "missing" sentinel was wrong. Plausibly explains the benign "Pool timeout" noise already logged in `UPGRADE_PROGRESS.md`.
- **`react-native-document-picker`** — confirmed via npm that this package is dead upstream (deprecated, renamed to `@react-native-documents/picker`), so the patch is the only way to keep it running on RN 0.81, not a stopgap for a fix that's coming.
- **credo-ts patches** (VCDM 2.0 shim, DI cryptosuite matching, biometrics-backed key option) — each additive-only and confirmed load-bearing against real call sites (`bc-agent-modules.ts`, `agent.ts` construct `SecureEnvironmentKeyManagementService({ biometricsBacked })`; without the patch this wouldn't typecheck).
- **`@animo-id/pex`/`@sphereon/pex`** — confirmed byte-for-byte unchanged in this PR, just carried forward from before the upgrade (matches `UPGRADE_PROGRESS.md`'s own note that upstream bifold 3.0.16 carries the same patches).

### Maintainability breakdown — three different categories, one real action item

**1. Bug-fix patches (RN TextInput, indy-vdr sentinel) — low ongoing risk.** These will eventually be obviated by an upstream release. Worth noting explicitly: yarn's `patch:` protocol fails the install (not silently) if a patch can't apply cleanly to a bumped version, so there's already a built-in forcing function here — when RN/indy-vdr are next bumped and one of these patches fails to apply, that's the trigger to check whether upstream picked up the fix and drop the patch. No separate tracking needed, just don't blindly regenerate a failing patch without checking upstream first.

**2. Credo-ts protocol-extension patches (`core`, `didcomm`, `react-native`) — recurring cost until upstreamed.** Unlike category 1, these aren't upstream bugs — they're Keyring-specific features (VCDM 2.0, DI cryptosuite matching, biometrics-backed keys) that credo-ts doesn't natively support. There's no guarantee upstream ever adds them, so **every future credo-ts bump (0.6.4, 0.7.x, ...) carries a real chance these need manual rework**, not just a mechanical reapply. Worth budgeting for that explicitly on the next credo-ts bump, and worth considering whether the VCDM 2.0 shim in particular is generic enough to propose upstream (reduces this repo's carrying cost if credo-ts adopts it natively).

**3. `react-native-document-picker` — the one patch with no expiration date.** This is different in kind: the package is dead upstream, so this patch has no path to ever becoming droppable on its own. It will need to be manually carried and re-verified against every future RN bump indefinitely, unless someone migrates to the replacement package (`@react-native-documents/picker`). Recommend filing this as its own tracked follow-up issue with an owner — indefinitely-patched dead dependencies are exactly the kind of thing that quietly turns into a security/compatibility liability a few years out, since the underlying package gets zero further upstream attention.

### One doc-drift finding, easy to fix, easy to forget

`docs/CRYPTO_SUITE_FOLLOWUP.md` (last touched 2026-07-13) still describes `DataIntegrityProof`/`eddsa-rdfc-2022` support as "intentionally deferred," and the `@credo-ts/didcomm` patch's own code comment cites "docs/CRYPTO_SUITE_FOLLOWUP.md, Decision 5 Option B" — but DI support actually shipped four days later (commit `1992195`), and there's no "Decision 5" in the current doc. Not a code problem, but a future reader following that comment's citation will find nothing. Worth a quick refresh of `CRYPTO_SUITE_FOLLOWUP.md` (and the corresponding `UPGRADE_PROGRESS.md` Phase 5 status text) to match what actually shipped.

### Suggested follow-ups (non-blocking on this PR)

1. File a tracked issue for migrating off `react-native-document-picker` to `@react-native-documents/picker` — this is the one item with a real deadline risk rather than a "wait and see."
2. Refresh `CRYPTO_SUITE_FOLLOWUP.md`'s "intentionally deferred" language and add the missing "Decision 5" entry the didcomm patch comment refers to.
3. Consider whether the VCDM 2.0 credo-ts/core shim is worth proposing upstream, to reduce future patch-carrying cost.

bmiller59:
**Missing `upgrade-baseline-p0` tag blocks the store-migration E2E test**

`UPGRADE_PROGRESS.md` (lines 63, 145, 386, 508, 550) and `e2e/README.md:247` both reference `upgrade-baseline-p0` as the Phase-0 baseline tag, created "in BOTH repos," and treat it as "the source of truth" for building the pre-upgrade baseline APK used by `yarn e2e:migration` / `yarn e2e:migration:android-only`.

That tag doesn't exist anywhere I can reach:

| Repo | Remote | Tags found | `upgrade-baseline-p0`? |
|---|---|---|---|
| keyring-wallet | `origin` | 1 | No |
| keyring-wallet | `AdvancedIdentity` | 14 | No |
| bifold | `origin` | 1 | No |
| bifold | `asml-bifold` | 5 | No |

Either it was created locally and never pushed, or it's been deleted since. This blocks re-running the Askar 0.2→0.6 store-migration E2E test as part of reviewing this PR. @albertoleon7794 — do you still have it locally (or know the actual commit), so it can be pushed/recreated?

bmiller59:
**Hardware-attestation toggle tap doesn't register on Android 16 (SDK 36) real device — `yarn e2e:vrc:witnessed:android-only`**

Ran the witnessed real-device VRC exchange on two physical Android phones:

| Phone | Android | SDK | Result |
|---|---|---|---|
| Samsung (older) | 13 | 33 | Hardware-attestation toggle worked, flow proceeded |
| Samsung (newer) | 16 | 36 | Toggle tap silently no-ops |

Failed identically on two consecutive runs, both times at the same step: `enableHardwareAttestation` taps the `ToggleHardwareAttestation` Switch, then waits for the resulting PIN screen → `Continue` button, which times out (`element testID=Continue not found in 15000ms`).

Evidence gathered from the failing device's page-source dump + logcat:
- The Switch element is legitimately present and interactable per the accessibility tree (`checkable="true" clickable="true" enabled="true"`), `checked` stays `false` after the tap
- Zero app-level (ReactNativeJS) log activity for the tap — it never reaches the attestation-toggle handler in JS
- No OS-level security block (Restricted Settings, tapjacking, AppOps denial) in logcat either

So the Appium command almost certainly executes, but something between UiAutomator2 and this specific (very new) Android build swallows the tap before it becomes a real touch event the app sees. Same device also needed a workaround earlier for `INSTALL_FAILED_VERIFICATION_FAILURE` installing Appium's UiAutomator2 helper APK — possibly the same general category of newer-Android-hardening-vs-automation friction, though I haven't confirmed they share a root cause.

Not related to the async-record-provider fix already in this PR — that fix's own e2e coverage (emulator-based `vrc-exchange:android-only`) passed clean. This is specifically the real-device attestation-toggle interaction on SDK 36.

Flagging as a known gap rather than digging further right now — let me know if you'd like it investigated further (e.g. coordinate-based tap as a workaround, or pinning a newer UiAutomator2 driver version).

bmiller59:
**Follow-up on the Android 16 attestation-toggle issue — traced further, found a likely deeper cause, still open**

Continued investigating the `yarn e2e:vrc:witnessed:android-only` failures from the previous comment, using the same two physical phones. Summary of what was tried and where it landed.

## What we fixed

Added a general `tapTestIdReliable(driver, key, verify, options)` helper to `e2e/lib/driver.js`: taps an element, then confirms the tap actually took effect via a caller-supplied `verify()` check, re-tapping (default 3x) if not. A plain WebDriver click reports success even when the app never receives the touch, so a single tap has no way to detect that — this does.

Applied it to four spots in `e2e/lib/flows.js` that showed a tap-succeeds-but-nothing-happens symptom over successive runs:
- `enableHardwareAttestation`'s tap on the `ToggleHardwareAttestation` Switch (verify: PIN screen appears)
- The `Continue` tap after PIN entry in that same flow (verify: PIN modal closes)
- `returnToContacts`'s tap on the `Contacts` tab (verify: leaves the current stacked screen)
- `openContactDetail`'s equivalent tap

Each fix measurably helped — the run got one step further each time (toggle → Continue → Contacts-tab), which is what led to the next finding.

## What's still open — looks like a deeper concurrency issue, not device flakiness

The fixes above patch individual symptoms, but the actual root cause looks more fundamental. Tracing one failure at the WebDriver-command level (not just the app UI) turned up this sequence on one phone's session (`24aa6832...`):

1. `findElement("Continue")` succeeds — element found, on the correct session.
2. The very next command, `getElementTagName` for that *same element*, is sent to a **different session ID** (`30bf9091...` — the *other* phone's session). Fails with "no such element" (unsurprising, wrong session).
3. Immediately after, queries back on the correct session for `Continue` come up empty — the screen had moved on.

`enableHardwareAttestation(sessionA)` and `enableHardwareAttestation(sessionB)` (and most other flows in this suite — onboarding, credential-offer acceptance, etc.) run concurrently via `Promise.all` against two independent WebDriver sessions in one Node process. The cross-session command above suggests WebdriverIO's client-side internals aren't fully isolating state between concurrent `remote()` instances — not something in our own flow code (the `driver`/`pinInput` references are properly scoped per call), and notably: this hit both the newer *and* the older phone across different runs, not just the one we initially suspected of being flaky.

If that's the actual mechanism, it would mean **any** concurrent two-device step in this suite is exposed to occasional cross-talk, not just the attestation toggle — which is a bigger question than a few tap retries can fix.

## Status

- The actual bug this session was chasing (the async-record-provider race in `@bifold/react-hooks`, first comment) is fixed and solidly verified via the emulator-based `yarn e2e:vrc:android-only` run.
- The `tapTestIdReliable` changes are applied locally (`e2e/lib/driver.js`, `e2e/lib/flows.js`), not yet committed/pushed — happy to push them regardless since they're a real improvement, but they don't fully resolve the underlying issue.
- Haven't yet confirmed whether this is a known WebdriverIO issue, a version-specific bug, or something fixable with better session isolation (e.g. avoiding shared globals, or checking if a newer/older webdriverio version behaves differently). Flagging for anyone with more context on this suite's concurrency model, and as something to pick back up with fresh eyes.

albertoleon7794:
**Android 16 attestation-toggle failure: root-caused and fixed — it was the harness's own `hideKeyboard`, not the tap and not session cross-talk**

Followed up on the `yarn e2e:vrc:witnessed:android-only` failures from the last two comments. I was able to reproduce the exact failure (`element testID=Continue not found in 15000ms`) **deterministically** on two clean Android 16 / SDK 36 emulators (pixel_7_pro profile, same panel class as the failing S25+), which made it bisectable. What actually happens:

1. The toggle tap **works** — the PIN modal opens every time. (The `checked="false"` in the page-source dump is expected at that point: `ToggleHardwareAttestation`'s state only flips after PIN auth completes, and `handleToggle` logs nothing, so "zero ReactNativeJS activity" doesn't indicate a dropped tap.)
2. Appium types the 6-digit PIN; the app dismisses the keyboard itself on the 6th digit (`PINInput` → `Keyboard.dismiss()`).
3. The flow then calls `driver.hideKeyboard()`. Appium's Android implementation checks `dumpsys input_method` and — racing the IME teardown animation — concludes the keyboard is still visible, so it sends `KEYCODE_ESC` (fallback `KEYCODE_BACK`). Caught the literal `adb shell input keyevent 111` in the appium log.
4. An RN `Modal` is an `android.app.Dialog`, and ESC/BACK cancels it. The PIN modal closes **unauthenticated**, the toggle stays off, and 15s later the harness reports `Continue not found` — with the failure screenshot showing the Secure Exchanges screen, no modal, toggle off.

A/B proof on a single emulator: identical steps *without* `hideKeyboard` → Continue found, tapped, toggle verified `checked=true`; *with* `hideKeyboard` → modal and Continue both gone. Since this reproduces with **one session and zero concurrency**, I don't think the wdio cross-session theory holds — what the command trace showed is wdio v9's `isExisting()` internals: on an already-found element it issues `getElementTagName(elementId)` and *expects* "no such element" as its normal false result (that's the verify-polling in `tapTestIdReliable` working as designed). Two sessions running the identical flow in lockstep also emit near-identical element IDs (UiAutomator2 uses per-session counters), so interleaved appium.log lines read like cross-talk when they aren't.

Why only the Android 16 phone: the bug is a race with the keyboard's self-dismissal, and it only fires if dumpsys still reports the IME visible when Appium checks. Android 13 / API 33 report it hidden fast enough that the keyevent is never sent; Android 16's IME timing lands the check inside the window every time.

Fix in 220b9de: `enableHardwareAttestation` now uses a `waitForKeyboardGone()` poll (no key events) instead of `hideKeyboard()`. The other `hideKeyboard` call sites are on plain, non-modal screens where ESC/BACK is harmless and are unchanged; your `tapTestIdReliable` helpers stay — they're good resilience regardless. Validated with the full concurrent two-device toggle flow on two API 36 emulators: both devices end `checked=true`. `yarn e2e:vrc:witnessed:android-only` on your Android 16 phone should now get past that step — would you mind re-running when you get a chance?

bmiller59:
### E2E: witnessed android-only real-device run — now passing ✅

Re-ran `yarn e2e:vrc:witnessed:android-only` end-to-end on two physical Android phones (serials `R5CN70Q6PDP` and `R5CY83SM4ST`) and it now gets all the way through:

```
============================================================
✅  E2E PASSED — vrc-exchange:witnessed:android-only
============================================================
```

Both wallets completed the witnessed + hardware-attested VRC exchange (`DataIntegrityProof`/`eddsa-rdfc-2022`), each ending up with a Witness Record.

This was preceded by a batch of e2e fixes/additions that got the run to a clean pass:

- `56252c4` feat(e2e): add android-only variants for migration and witnessed exchange tests
- `b13bae5` feat(e2e): print a clear pass/fail banner after every runner script
- `d8c43c2` fix(e2e): add a verify-and-retry helper for real-device tap flakiness
- `220b9de` fix(e2e): stop hideKeyboard from dismissing the attestation PIN modal on Android 16
- `918cfed` feat(e2e): add android-only variant of the real-device VRC exchange (non-witnessed)
- `b0f2894` fix(e2e): make returnToContacts survive the witness's message burst
- `2234fbd` fix(e2e): stop the witness server from orphaning its own process
- `90a7b62` fix(e2e): preflight both phones' lock screen before a witnessed run
- `4102e8c` docs(attestation): record real-device evidence of the expired legacy root
- `01ed818` fix(e2e): tolerate expired-legacy-root attestation as a warning, not a failure

**Known issue (tolerated, not a failure):** one side ("Alice") showed "Witnessed" without the "Secure Exchange" badge — the peer's chain validation failed with a "Hardware Verification Issue" because that phone's attestation chain terminates at Google's legacy attestation root, which expired 2026-05-24 (`GoogleAttestationRoots.kt`). This is a real device-population limitation, not a regression — see `docs/HARDWARE_ATTESTATION_FLOW.md` for details. The other side ("Bob") showed both "Witnessed" and "Secure Exchange" cleanly.

bmiller59:
### E2E: real-device android-only VRC exchange (non-witnessed) — passing ✅

Also ran the new `yarn e2e:vrc:devices:android-only` (two physical Android phones, no witness server in the loop) on the same pair — `R5CN70Q6PDP` and `R5CY83SM4ST`:

```
============================================================
✅  E2E PASSED — vrc-exchange:devices:android-only
============================================================
```

Same pattern as the witnessed run: `R5CN70Q6PDP` showed the full "Secure Exchange" banner (peer hardware attestation chain-validated), while `R5CY83SM4ST` showed the "Hardware Verification Issue" warning — the same known, tolerated limitation (peer evidence present, but that phone's attestation chain terminates at Google's now-expired legacy attestation root; see `docs/HARDWARE_ATTESTATION_FLOW.md`). Not a regression, and the run correctly continues rather than failing on it. Both wallets ended up with each other's VRC.

---

### Review complete

That closes out review on this PR. Non-blocking items I called out above (DRY follow-ups in the VRC module, `vrc-reference` capability-negotiation coverage, the RCard cross-credential composition path, `witness-server` test coverage for the dual-stack behavior, the dead-end left in `app/src/services/attestation.ts` by the DRPC removal, E2E harness doc gaps, credential-versioning/type-detection consolidation, and the dependency-patch maintainability notes) will be converted into follow-up tickets rather than block merge here. The Android 16 attestation-toggle bug is already root-caused and fixed (thanks @albertoleon7794). The missing `upgrade-baseline-p0` tag is still open and blocks only the store-migration E2E test specifically, not this review.

bmiller59 (approved):
Approved! Great work. I had to resolve a few quirks with my particular devices but code looks good and all tests and e2e tests are passing over emulators and my two android devices.

See in the PR comments for some small follow up issues.

albertoleon7794:
**Review follow-ups: all addressed on this branch**

Thanks for the thorough review. Rather than converting the non-blocking items into follow-up tickets, everything has been addressed directly on this branch:

- **Witness-server test coverage** (`7ef07bbd`) — the freshness/digest/VWC logic is now exported and the unit suite tests the real implementation (the old suite's only import was a type, so jest never actually loaded `WitnessService`); includes the 10-minute-window pin, JCS digest invariance, and the VC1.1/VC2.0/DI mirroring branches.
- **Display architecture + type detection** (`c68546a5`) — canonical `credentialTypes.ts` predicates replace the 15 inline type checks; the handler registry gained `extractSubject` + cross-credential context, so the CredentialOffer header/field-list drift is fixed and structurally prevented; new version-compatibility matrix test covers `{v1, v2, v3} × {±RCard}` plus the header/fields agreement invariant.
- **DRY follow-ups** (`08c90af7`) — single `counterpartyRceVersionAtLeast` helper, shared `selectCredentialContexts`, exported `toRawCredential` replacing the nine inline `record.encoded` reads.
- **`app/services/attestation.ts` dead-end** (`5714020`) — resolved by removal: Keyring doesn't consume Traction-gated credentials, so the monitor, restrictions config, and the notification-hiding filter are gone. A BC-style proof request would now surface as a normal, visible request. Core's default (no monitor registered) applies, so this is a stock upstream configuration. Person-flow is deliberately untouched — separate decision.
- **`vrc-reference` scope** (`5042fb96`) — took the documentation route: `BIFOLD_FLOW_ALIGNMENT.md` now states the negotiation/DI/RCard scope limitation prominently; full parity is planned as part of turning the package into a proper conformance suite.
- **Docs sweep** (`b2aed87a`, `f0cbd84`) — locality-bypass caveat in the runbook and e2e README, `debug-witness-connect.js` documented, `CRYPTO_SUITE_FOLLOWUP.md` refreshed to the shipped-DI state (including the Decision 5 record the didcomm patch cites), document-picker migration tracked in `UPGRADE_PROGRESS.md`.
- **`upgrade-baseline-p0`** — existed locally but was never pushed; now on both origins, so the store-migration e2e is unblocked.

**Two reminders for merge time:**
1. This PR carries the Google attestation trust-anchor fix (`88f80fe8` — the root rotation plus the self-as-anchor chain-validation vulnerability you called out). Merging this PR is what ships it.
2. Once merged, keyring-bifold #20 and keyring-wallet #3 (the root-rotation tracking issues) should be closed — the fix supersedes them.

PR #18, opened 2026-08-19
Locality: co-presence evidence a third party can trust, ref-06p, and the proof-set decision

bmiller59:
@albertoleon7794 — two things in here touch your work directly, flagged up front:

1. **`CRYPTO_SUITE_FOLLOWUP.md` Decisions 3, 6 and 7 are superseded** by a `eddsa-jcs-2022` + `bbs-2023` proof set. That is your cred-spec #18 recommendation applied to us, so this is mostly ratifying your own proposal on the Keyring side — but it reverses a decision we made together in July and it deserves your eyes.
2. **A comment went onto cred-spec #18** ([link](https://github.com/trustoverip/dtgwg-cred-spec/pull/18#issuecomment-5336345589)) supporting the JCS recommendation with a third argument. Brendan reviewed it before it went up and has told you separately; noting it here so it is in the record too.

---

## What this is

A new plan — **locality**: co-presence evidence a third party can trust. A witness deployed at a venue observes the parties' devices over a short-range radio channel it operates, and attests to that observation inside the VWC it already issues. Plus its first reference rung, and the cryptosuite decisions that fell out of designing it.

Built on the OpenVTC plan's Trust Task recast, not on the legacy chat ceremony — every seam it needs (`ext`, the §4.9.3 task digest, the retained outcome pair) is a task-layer facility, so this is a consumer of that work rather than a parallel track.

## The four commits

| | |
|---|---|
| `c935c97` | `chore(openvtc)`: pin `dtgwg-cred-spec` at `b89f389` |
| `0d7fb8b` | `docs(plans)`: the locality plan + `ref-06p` |
| `6aad847` | `docs`: adopt the `eddsa-jcs-2022` + `bbs-2023` proof set |
| `4089905` | `docs`: raise the argument on #18, file the attestation bug |

## The design, in short

- **The witness observes; the device does not report.** Neither platform will attest a radio observation, so a device's "I heard the beacon" can never be third-party evidence. This **inverts the direction of the `LocalityService` scaffold already in the tree**, whose broadcast challenge is a shared secret with a five-minute life that any passive listener can hand to anyone on earth.
- **The advert LOCATES, the signed GATT transcript BINDS** — the same split `ref-06w3` forced at the document layer, one layer down at the radio.
- **It rides in `ext`** under `edu.harvard.seas.atl.keyring` on the four merged `witness/*` documents, so **no upstream spec change is needed** and a locality-blind peer degrades by the framework's own MUST-ignore rule.
- **Evidence is typed, negatives explicit.** `confirmed: false` with a reason, an absent member, and a confirmed assertion are three states a verifier can tell apart. Annotate by default; gating is a per-event deployment choice.
- **Honest scoping, at the design rather than in a footnote:** the claim is venue-scale, not room-scale, and the timing bound excludes long-haul relays and nothing finer. Witness-free peer-to-peer locality is **closed, not deferred** — the observer set equals the interested set, so no version of it produces third-party evidence.

## The rung — `ref-06p`, 26 checks, no radios

Four forgeries each rejected by a named check (replay across sessions, EID copy, counterfeit session document reusing an `id`, transcript lifted from another exchange), `ext` across the published `@openvtc/trust-tasks` 0.9.0 §7.2 pipeline, and **1,887 bytes/session measured**.

It also found a **live defect**: the credential's JSON-LD context defines `localityVerification` but none of the members nested inside it, so under RDF canonicalization today's locality evidence would be unsigned — safe mode refuses to canonicalize, unsafe mode drops it to zero quads while it still sits in the JSON. Latent only because the sole provider is `NullLocalityProvider`.

## The cryptosuite reversal, and why

ZKP presentation is a requirement, not an aspiration — the spec we edit says implementations SHOULD make it the default. That makes `bbs-2023` load-bearing, and `bbs-2023` is RDF-canonicalized, so **RDF does not leave the codebase**. What does leave: `Ed25519Signature2018` and `eddsa-rdfc-2022` issuance, and the `rceVersion` negotiation ladder.

Decision 7 kept dual-verify forever because *"stored credentials make that verify path effectively permanent"*. That is false for this credential type: `DEFAULT_CREDENTIAL_EXPIRATION_DAYS = 7` in both the app and the witness server, so the population self-clears a week after issuance stops. Pre-production on top of that.

**"Keep RDF for `bbs-2023`" is not "keep the legacy suites"** — those are two decisions and they were being confused.

Decision 13 makes the JSON-LD vocabulary a **CI guard rather than a discipline**, because we have already failed that discipline once, silently, in shipped code.

## Two things deliberately not here

- **No BBS+ tooling survey.** Deferred on purpose: a second proof is additive, a credential's **member layout is not**, and the layout is fixed without any of it. The assertion is already flat, disclosure-tiered, and carries predicates rather than identifiers — the device key id stays on the artifact side, because a stable key id across ceremonies is a correlation vector. Open as §11-Q5/Q6.
- **No bifold change.** The `useHardwareAttestation` split default is filed as [keyring-bifold#38](https://github.com/berkmancenter/keyring-bifold/issues/38) instead; the submodule pointer is untouched.

## Where to look

- Design → [`docs/plans/locality-plan.md`](https://github.com/berkmancenter/keyring-wallet/blob/plan/locality/docs/plans/locality-plan.md)
- **Reasoning, tradeoffs and superseded positions** → [`docs/plans/locality-plan/2026-08-18-bam.md`](https://github.com/berkmancenter/keyring-wallet/blob/plan/locality/docs/plans/locality-plan/2026-08-18-bam.md) — 33 findings, including several correcting my own earlier positions in the same session
- Suite decisions → [`docs/CRYPTO_SUITE_FOLLOWUP.md`](https://github.com/berkmancenter/keyring-wallet/blob/plan/locality/docs/CRYPTO_SUITE_FOLLOWUP.md) "Superseding decisions"
- Run the rung → `cd tsp-reference/ref-06p-locality-binding && npm install && npm start`

## Open questions in the plan

The venue machine and its BLE adapter (Q1), whether the first event gates or annotates (Q2), NFC kiosk scheduling and Apple's HCE entitlement (Q3), venue-claim provenance (Q4), and the deferred BBS+ tooling and issuer BLS key (Q5/Q6).


albertoleon7794:
Closing this rather than merging it: the content is already on `main`.

It landed via `f959433` outside this PR. I compared all 20 files this branch touches — **19 are byte-identical to main**, including `locality-plan.md` (same blob, 77,612 bytes) and `CRYPTO_SUITE_FOLLOWUP.md` with the `eddsa-jcs-2022` + `bbs-2023` proof set.

The 20th difference runs the wrong way: `main`'s `docs/plans/README.md` has the `release-flow-plan.md` row (from #22) that this branch predates, so **merging would delete it**. That is also why GitHub shows this as conflicting — same content, different commits.

The review exchange is complete and preserved on main: `2026-08-19-al.md` (three findings) and `2026-08-19-bam.md` (all three adopted, with `windowSeconds` narrowed after checking `ref-06p/run.mjs`).

Nothing is lost. Follow-ups belong in new dated companions under `docs/plans/locality-plan/`, per `docs/plans/CLAUDE.md`.

Separately, and worth noting here for the record: the cred-spec split this PR's review discussed was blocked on `dtgwg-vti-spec` not existing. All six DTGWG spec repos now exist as of 2026-09-04, so that blocker is gone.

PR #20, opened 2026-08-21
The relationship exchange and witness ceremony on Trust Tasks (steps 5–7 of the plan) — proven on devices

albertoleon7794:
**Branch:** `feat/trust-tasks-integration` → `main` (32 commits). Pairs with keyring-bifold PR berkmancenter/keyring-bifold#39 (`feat/trust-tasks-integration`, 18 commits) — the submodule pointer in this branch tracks it; merge the bifold PR first, then this one.

## What this is

The VRC/witness recast onto Trust Tasks, per `docs/plans/openvtc-integration-plan.md` and its subtask `openvtc-integration-plan/trust_tasks_subtask.md` (§9 steps 5, 6, 7). For two current (v4) wallets the whole relationship exchange now runs as Trust Task documents over the binding-0.2 DIDComm carriage:

- **discovery → propose → one-tap consent** (the relationship proposal is the consent moment; there are no per-credential offers any more),
- **hardware-attested signed issue legs** with digest receipts both ways, verified under the sender's relationship DID (eddsa-jcs-2022 document proofs),
- **per-party witness sessions** (challenge-bound VPs; VWCs bound by `taskContext` + `taskDigestMultibase`; outcome evidence retained),
- **presentation assembly + the verifier's pairing algorithm** (cred-spec Outcome Interpretability),
- **witness-share**: each wallet ships its VWC bundle to the peer, who runs the full pairing before lighting the **Witnessed** indicator — earned by verification, not granted on receipt.

**The flow, end to end (two current wallets and a witness):**

```mermaid
sequenceDiagram
  participant A as Wallet A
  participant W as Witness
  participant B as Wallet B
  A->>B: trust-task-discovery
  B-->>A: supportedTypes
  A->>B: propose relationship
  Note over B: one-tap consent
  B-->>A: propose response
  par each wallet runs its own witness session
    A->>W: witness session
    W-->>A: challenge
    A->>W: submit presentation
    W-->>A: VWC bound to the session
  and
    B->>W: witness session
    W-->>B: challenge
    B->>W: submit presentation
    W-->>B: VWC bound to the session
  end
  A->>B: issue signed VRC
  B-->>A: digest receipt
  B->>A: issue signed VRC
  A-->>B: digest receipt
  A->>B: witness-share VWC bundle
  B-->>A: receipt after verifying
  B->>A: witness-share VWC bundle
  A-->>B: receipt after verifying
  Note over A,B: Witnessed badge lights on both contacts
```

Legacy peers (v3) keep the untouched legacy flow (capability-gated; `rceVersion` stays the bootstrap beneath discovery). The R-Card (contact card VDS) still travels the legacy issue-credential leg — now also triggered by the accepted proposal so a lost announcement can't starve it.

## Evidence

- Witnessed two-wallet e2e on the **production mediator**, simulator + emulator: green (latest runs 36–37 after the final refactor). Gates = run-scoped logcat markers (exchange, ceremony incl. the outcome-evidence self-check, witness-share round trip) + UI (contact name, Witnessed badge on both screens).
- **Attended real-device runs** (Samsung + iPhone): green, repeatedly — hardware-attestation evidence present (4-cert chain), **Witnessed + Secure Exchange on both phones**. Demo runbook: `docs/DEMO_RUNBOOK_WITNESSED_EXCHANGE.md`.
- Unit: `bifold/packages/core` trust-tasks + vrc suites (223), witness-server suite (617).

## Also in here

**Where the code lives now (one plumbing, one runtime, both sides):**

```mermaid
flowchart TB
  app[Keyring app] --> core[bifold core - trust-tasks module]
  core --> shared[bifold trust-tasks - shared plumbing]
  witness[witness-server] --> shared
  core --> runtime[openvtc trust-tasks runtime 0.9]
  witness --> runtime
  shared -.- what[document proofs and digests, carriage message, payload validator]
```

- `@bifold/trust-tasks` — new platform-neutral package: the document-proof/digest, carriage-message and payload-validator code the wallet and the witness-server both run (retires two deliberate duplicates; `vrc-shared` stays server-only per its README). App wiring = portal + app dep + Metro entries (recipe in `CLAUDE.md`).
- The witness-server consumes through the real `@openvtc/trust-tasks` runtime (ESM-only, loaded via dynamic import); its local pipeline is deleted.
- e2e: the sim witnessed runner (`yarn e2e:vrc:witnessed`), the devices flow rewritten to the v4 UX, marker/UI gates, README section "The Trust Task dialect (v4)".
- Transport hardening from the connection-debug work: retrying HTTP outbound transport, PickUpV2 polling, the didexchange watchdog.
- Plan at Rev 15; dated companions `2026-08-18-al` → `2026-08-20-al` carry the reasoning and the e2e ladder's findings; `docs/spikes/` has the evidence notes.

## Reviewing this

Start with `bifold/packages/core/src/modules/trust-tasks/ceremony.ts` (the exchange engine), then `witnessCeremony.ts`, `outcomeEvidence.ts`, `witnessShareSpec.ts`; witness side `witness-server/src/trustTasks/WitnessTaskSessions.ts` + `runtime.ts`; shared `packages/trust-tasks/src`. UX: `RelationshipProposalModal`, the overlay wording in `hooks/chat-messages.tsx`, `vrc-manager.ts` (`issueRCardForAcceptedExchange`, the offer hold).

After pulling: `yarn install` at the root, `git submodule update --init`, `yarn start --reset-cache`; rebuild the iOS device app before any device run (it bundles its JS).

## Known gaps / not in this PR

- v3-compat (old wallet ↔ new) is true by construction, not yet evidenced against an old build.
- Proof-set migration (eddsa-jcs-2022 + bbs-2023, Decisions 10–13) deliberately parked pending the working group (cred-tf discussion #39); VRC/VWC credential proofs still sign eddsa-rdfc-2022, task documents eddsa-jcs-2022.
- Simulator-only intermittents (witness-connect miss, iOS no-send) documented in the e2e README; none seen on devices.
- Speed: ~60–90 s per exchange is mostly 10 s pickup polling; adaptive polling is a follow-up.
- R-Card as its own Trust Task — open design question for the group (recommendation recorded).


bmiller59:
Did a full-branch code review of this PR (`main..HEAD`, 59 commits), plus a follow-up pass. The core Trust Task wiring (`useBCAgentSetup.ts`, `bc-agent-modules.ts`) is correct and well-justified; what turned up was mostly in the e2e harness:

- `e2e/lib/witnessedExchangeFlow.js` (and 4 other runner call sites) — the bidirectional `acceptRelationshipProposalIfPrompted(a)` / `(b)` calls had no check that *either* side actually saw the proposal. If discovery failed and neither device was prompted, both calls silently returned `false` and the run fell through into `assertVrcReceived`'s generic "contact never appeared" timeout instead of a clear "no proposal shown" failure.
- `app/metro.config.js` — the `@openvtc/trust-tasks/*` resolver silently fell through to Metro's default resolution when no `dist/<subPath>.js` matched, reproducing the exact "no match was resolved" failure it exists to avoid.
- `e2e/lib/flows.js` — `openContactDetail` treated any visible `ContactMenu` as "already on this contact's chat" without checking the peer's name was actually on screen — a caller retrying after a failed match could silently assert against the wrong contact.
- `app/src/hooks/useBCAgentSetup.ts` — `RetryingHttpOutboundTransport` retried on *any* exception, not just the documented stale-socket case, risking a silent duplicate send of a non-idempotent DIDComm message.
- `e2e/lib/config.js` — `resolveWitnessMediatorInvitationUrl()` doesn't strip quotes or inline comments from `MEDIATOR_URL=` in `app/.env`, so a quoted value would produce a malformed invitation URL. Flagging but not fixed — lower priority, didn't want to touch more of the harness than the mediator investigation already needed.

Fixed the first four in #25 (`fix/mediator-pickup-strategy` → `feat/trust-tasks-integration`, 9 commits, `yarn typecheck`/`yarn lint` green). That branch also carries the actual mediator-pickup deadlock fix (explicit PickUpV2 polling, the reason this review happened) plus the bifold-side R-Card overlay fix (keyring-bifold#41).

Merge order: keyring-bifold#41 → #39, then #25 → this PR — matching the pairing note in the description above.

bmiller59:
Also worth flagging on the wallet side, since it's the same root cause: the mediator-pickup bug fixed in #25 was silent — no error surfaced on either end, and the wallet UI would even show "You connected and can now message securely" on a half-open handshake, because `chat-messages.tsx` renders that banner whenever a connection *record* exists, with no state check. The effective pickup strategy also isn't fully determined by module config: credo prefers whatever strategy is persisted on a wallet's `MediationRecord`, so the same code could look fine on one wallet and fail deterministically on a fresh one — which is why this survived undetected from 2026-08-24 until now.

#25 confirms the fix on real hardware in both DIRECT and MEDIATOR transport modes (2026-09-01) — the first time mediator-mode witnessing has been proven working end to end on physical devices rather than by unit test or code inspection. Full diagnosis chain and evidence in [`docs/spikes/e2e-vrc-connect-findings.md`](https://github.com/berkmancenter/keyring-wallet/blob/02a4cf603996359e2bcc8d35a07b8d86fc299c01/docs/spikes/e2e-vrc-connect-findings.md#fourth-failure-layer-the-witness-never-picks-up-its-mail-2026-08-31-bam), "Fourth failure layer: the witness never picks up its mail" (plus the earlier stale-socket and live-mode-pickup-loss layers above it in the same file, which the mediator work in this branch also carries fixes for).

PR #24, opened 2026-08-31
docs(plans): packaging Keyring as a reference app + SDK — working notes

bmiller59:
## Summary
- Exploratory checklist + per-idea analysis of packaging this repo as a reference app + SDK, so prospective developers can try distinct identity/credential use cases (The Approver, Passkeys, More Factors, The Agent Leash, Prove You're Human, Recovery, non-forwardable invites, The Quorum, Human Factor Authentication, and a trading-card exchange demo).
- Reconciled against `openvtc-integration-plan.md` §6's own "Demo concepts" brainstorm — cites the overlap explicitly (e.g. Agent Leash ≈ "Meet your Agent") rather than treating these as independent proposals, and corrects an early research pass that missed `bifold/packages/core/src/modules/trust-tasks/`'s already-generic consume/respond engine.
- Adds a "developer path" section, walked concretely (clone → mediator → build → theme → DI wiring → register content → exchange), with five requirements (R1–R5) derived from real friction found in that walkthrough — most notably the 418-line `AppContainer` having no trimmed starter template, and no local-mediator tooling equivalent to the witness spin-up script.
- Status: exploratory notes, not a commitment to implement — flagged in the doc itself, not yet split into a proper plan + dated companions per `docs/plans/CLAUDE.md`.

## Test plan
- [x] Docs-only change; no app code touched.
- [x] `yarn typecheck` / lint ran clean via the pre-commit hook.
- [x] Pre-push test suite ran clean (16 suites, 70 tests).

### DIFF
--- a/docs/plans/reference-app-sdk-packaging.md
+++ b/docs/plans/reference-app-sdk-packaging.md
@@ -6,6 +6,12 @@
 
 **Positioned as a sibling to the openvtc plan, not a subtask of it** — same relationship `locality-plan.md` has (openvtc-integration-plan.md §4.6: "the locality axis now has its own plan"). It depends on that plan's infra (Trust Tasks, witness recast, PNM/VTA) but its own scope — developer packaging, DI/theming ergonomics, a trading-card demo — extends past Trust Tasks into territory the parent plan doesn't own. It stays a standalone exploratory doc rather than moving under `openvtc-integration-plan/` because subtasks there state *current design* in present tense (per `docs/plans/CLAUDE.md`); this is still pre-decision brainstorming.
 
+**Review index.** Companions carrying the reasoning behind the positions below:
+
+| Companion | Author | What it settles |
+|---|---|---|
+| [`2026-09-01-al.md`](./reference-app-sdk-packaging/2026-09-01-al.md) | AL | SDK scope and framing (mobile-first, SDK as the deliverable); consume upstream Trust Tasks rather than reimplement, and the duplicate-implementation problem it exposes; wrap orchestration rather than types; the VTA Farm as first-class onboarding; `ref-08-pnm-core-hermes` as the unblocking experiment; requirements R6–R12 |
+
 **Reconciled against `openvtc-integration-plan.md` §6 ("Demo concepts — brainstorm only")**, which already proposes several of the same demos under different names, with more ecosystem grounding (real Trust Task type URIs, a stated flagship target). This file's per-idea notes below cite the overlap explicitly rather than treating these as independent proposals:
 
 | This file's idea | §6 prior art |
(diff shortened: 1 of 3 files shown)


albertoleon7794:
Pushed two commits onto this branch — your notes are untouched, both are additive.

**`3a09e34`** adds a dated companion with the SDK half stated from the product angle, and an *Adopted positions* section on the plan pointing at it. The substantive ones: Trust Tasks gets **consumed, not reimplemented** (`@openvtc/trust-tasks` is generated registry bindings plus a zero-dep runtime that *injects* crypto via `proofPolicy`/`payloadPolicy` — we own what it leaves blank, including hardware-backed signing); the package splits on **two** axes since platform-neutral isn't enough while `@bifold/trust-tasks` depends on Credo; and **R5 is restated** as `registerTrustTask({ spec, orchestration, renderer })`, because the framework is document-level and says nothing about what `ceremony.ts` already does — proposer selection, peer capability discovery, version gating, idempotence on redelivery. That orchestration is the SDK's product surface, and it's the reason wrapping earns its keep.

One finding in there is worth your eye independently of the packaging question: **`@bifold/trust-tasks` doesn't depend on the upstream package at all** — it's a parallel implementation over `canonicalize` + `ajv`. Since `pnm-core` *does* depend on it, bringing PNM into the wallet would ship two Trust Tasks implementations, two JCS canonicalizers, in one bundle. That doesn't fail at build time; it fails as an intermittent per-document signature mismatch. The companion sequences it T0–T4, and T0 (advance the pins — upstream is at 0.16.8 against our 0.9.0) and T1 (measure whether the two canonicalizers actually diverge, before touching code) are independent of everything else here.

**`0d4f8c9`** carries the rest of the requirement set: R13–R17 (docs stack with the API-report gate that makes self-updating docs real, stable hosted address, `llms.txt`/`AGENTS.md`, diagrams; SDK conventions from Stripe/Twilio/Plaid; a builder intake channel; a showcase surface separate from docs; multi-language via published conformance vectors rather than OpenAPI generators, which can't carry the canonicalization and verification). Plus two acceptance criteria written so they can fail — the reference app consumes the *published* artifact, and "stupid easy" is a timed walkthrough by someone who hasn't seen the repo.

Two things that change your text rather than extend it, so flagging explicitly:

- **R1 gets reclassified.** Every mature SDK ships a sandbox; ours is the local stack plus the Farm. So the mediator friction you found isn't DX polish, it's table stakes — and upstream's own guidance is the managed [VTA Farm](https://vtafarm.firstperson.dev/) as Path A, with self-hosting as Path B (where `ref-05-local-vta` is already the precedent). R1 becomes "point at the Farm" rather than "hand them a compose file."
- **A critical-path experiment now sits ahead of the packaging work.** `@openvtc/pnm-core` carries a complete TypeScript VTA client under `src/vta/` on an RN-viable stack (noble, scure, cfworker/json-schema, cbor-x) that nonetheless mentions neither React Native nor Hermes anywhere. `ref-08-pnm-core-hermes` would answer three things this plan is currently guessing at: whether Keyring can talk to a Farm VTA, whether the Agent Leash is a port or a rewrite, and whether the SDK owns a VTA client or consumes one.

Also recorded, since you'd want to check before scheduling: `@openvtc/rp-sdk` already exists upstream for the relying-party surface, and `pnm-core` is described as a WebAuthn-passkeys-to-VTA bridge — substantially the mechanism behind the slate's passkey product.

Happy to split any of this out if you'd rather the plan stayed yours and this became a separate PR.

bmiller59:
@albertoleon7794 — pushed a follow-up commit (62d5a65) with the phasing decision this drove: the lab director wants spin-up-and-use demos in the short term for fundraising, which settles the reference-app-vs-SDK question for now — this phase is entirely R1–R5, R6–R17 stay deferred (not reversed).

New dated companion: [`2026-09-03-bm.md`](../blob/plan/reference-app-sdk-packaging/docs/plans/reference-app-sdk-packaging/2026-09-03-bm.md), cross-linked from the plan's review index and a new "Phasing and priorities" section.

Worth your review in particular:
- The one-app/additive-registry/picker architecture for the two demos (trading card, approval) — meant to double as a live instance of your `registerTrustTask({ spec, orchestration, renderer })` proposal from §2.
- Hardware-signing decoupling pulled forward in the sequencing, since it's dual-purpose (approval demo + your own extraction list).
- The cross-check against `feat/trust-tasks-over-didcomm-v1`: `main`'s `ceremony.ts` still calls `agent.modules.didcomm` directly throughout, but that branch already has a `Carriage` port abstracting exactly that away — real evidence your T3 (pure/Credo-adapter split) is cheaper than the companion assumed. R5 is now specified to be built Carriage-shaped from day one so it isn't reworked when that branch merges.
- The companion's last section lists what should be communicated to whoever owns that branch (file-contention heads-up, merge-timing ask, telling them their work de-risks T3, type-URI namespacing) — flagging in case you're the right person to relay that, or know who is.

Only the T1 canonicalizer measurement from your remediation list is pulled into this phase, as pre-demo-day insurance; T0/T2–T4 stay on the SDK track's own timeline.

PR #26, opened 2026-09-03
docs(plans): Keyring on the VTA Farm — the phone at every level of the developer path

albertoleon7794:
A new sibling plan, proposed for review, plus its dated companion and a row in the plans index.

**Based on `plan/reference-app-sdk-packaging` (#24)**, not on `main`, because it builds directly on that branch's `2026-09-01-al.md` companion and closes one of its open questions. Retarget to `main` if #24 lands first — the diff is three files and touches nothing #24 owns.

## What it proposes

Keyring as a full participant at every level of the developer path **upstream itself documents** (`vti-setup/developer/`, pinned clone), against a **Farm-hosted** VTA rather than one we run:

| | Upstream's step | Keyring's target |
|---|---|---|
| **L0** | Create a VTA in the Farm console, connect a local PNM | The phone displays its temp `did:key`; the human pastes it into the tab already open |
| **L1** | Bind the TUI, mint a persona | The wallet *is* the everyday interface |
| **L2** | Submit a join request under an M-DID | Join from the phone |
| **L3** | *"at least two valid VRCs whose issuers are both `Active` members"* | Those VRCs are Keyring-minted, witnessed, hardware-attested |

L3 is the point of the plan: everywhere else Keyring is catching up to a CLI, and there the ecosystem's membership gate consumes exactly the artifact we already produce better than anything else in the stack.

## Why it is a separate document

It is the first plan whose counterparty is infrastructure **we do not run**. A frozen fixture cannot stand in for the Farm agreeing with us; every blocker resolves to a workaround or an upstream PR; and it is the only plan that can retire "works against our own VTA" as the standard of evidence.

Scope is deliberately narrow — it owns only what changes when the VTA is someone else's managed service. The transport and operation layers stay with `openvtc-integration-plan.md`, the VTA client itself with `pnm_cnm_subtask.md`, and developer packaging with this PR's base.

## Three things the companion settles with measurements

- **Hardware custody, decided.** The Farm asks for a *client-generated* temp `did:key` and pasted grant, not `provision-integration`'s off-device key — so `pnm_cnm_subtask.md` §9 Q2 resolves in favour of the enclave key, on this path. Enrolment collapses rather than disappears: the out-of-band operator step is a paste, and the operator is the user.
- **No WebAuthn on the wallet.** The passkey belongs to the Farm *console*; the VTA session is `/auth/challenge` → authcrypt → bearer, and TSP/DIDComm carry no bearer at all. Closes the open question in `reference-app-sdk-packaging/2026-09-01-al.md` §4 and unblocks the Agent Leash framing.
- **The `did:webvh` chain is a packaging defect, not a design defect** — and choosing REST does not avoid it. `pnm-core` 0.7.0 → `vti-didcomm-js` 0.7.0 → `didwebvh-ts` 2.8.0, whose `react-native` export condition points at the build containing `require("node:fs")`. The fs access is lazy and guarded; only Metro's *static* resolution breaks. All reproducible from the npm registry — commands are in the companion.

## The contribution track

Seven candidates, evidence-backed, approval-gated as always. The two cheapest are measured rather than guessed: of the 12 files in `pnm-core` importing the `vti-didcomm-js` barrel (and **zero** importing a subpath), the dominant imports — `base64url` ×6, `multibase` ×5, `didKey` ×2 — all have clean dedicated subpaths that exist today; and `createResolver` already advertises a pluggable method map that its own static `did-webvh` import defeats. Two mechanical changes, no API change, no design argument.

## One new rung

`ref-16-farm-membership` — the terminal rung, deliberately last and deliberately one rung rather than four: it is the only rung whose failure cannot be localised, which is what an end-to-end proof *is*, and it is informative only once everything beneath it is green. Bisects against `ref-05-local-vta`, which becomes the offline CI twin.

## Worth arguing about

- **Sibling vs subtask.** Filed as a sibling on the locality/SDK precedent. It could equally be a third subtask under `openvtc-integration-plan/`.
- **The pin advance in F0.** All four npm tripwires have tripped and the trust-tasks clone is 68 commits behind `origin/main`; the HTTPS binding's endpoint-discovery contract is one of the things that moved. I have stated that as a gate rather than a fact — but it does mean F0 starts by moving pins mid-workstream.
- **L3 rests on a policy someone else controls.** The two-VRC gate is quoted as a target that *"is not what is currently active today"*. L0–L2 don't depend on it; the strategic argument does.
- **§9 Q3 is a question for upstream, not for us**: the join task URI in the guide uses the `trusttasks.org/openvtc/` authority that `pnm_cnm_subtask.md` §3.4 records as retired. One sentence resolves it; a day of implementation discovers it.


bmiller59 (commented):
Overall this is well-evidenced and I like the direction — the Farm materially lowers the barrier to entry over self-hosting, and the §4 packaging-defect finding on `didwebvh-ts`/`pnm-core` is a genuinely useful, independently-checkable measurement regardless of what we decide on the Farm itself.

One thing I want stated explicitly rather than left as an implicit reading of "sibling plan": **VTA Farm support should be valuable but fully optional, and must not become a core dependency of `openvtc-integration-plan.md` / `pnm_cnm_subtask.md` / `reference-app-sdk-packaging.md`.** We should be able to hit our core OpenVTC goals with or without Farm support ever shipping.

The plan's structure already mostly gets this right (sibling doc, `ref-16` appended at the end of the ladder rather than inserted into it, `ref-05-local-vta` kept as the CI default). The one place the *wording* drifts from that is §3.4: "This plan is what makes it urgent rather than eventual" reads as license to pull `did:webvh` resolution forward on the parent's roadmap because of the Farm. But that need isn't Farm-created — `ref-06x-cypress-stack` proved it against our own self-hosted VTA on 2026-08-17, and `pnm_cnm_subtask.md` already marks it "shared with parent Phase D item 3." So the accurate claim is narrower: the Farm increases *this plan's own* exposure to the adapter not existing yet, not the adapter's priority for the parent.

I pushed a commit (`96fed25`) that:
- adds a **Dependency direction** line to the header stating the non-core relationship plainly,
- adds a §1 paragraph making the constraint explicit with citations,
- reworks the §3.4 sentence above so the urgency claim is scoped correctly,
- flags §7's contribution track as spare-capacity work rather than phase-gating,
- and adds a dated companion (`2026-09-04-bm.md`) recording the reasoning, citations, and what it does/doesn't change (nothing in AL's §3/§4 measurements or adopted positions is revisited).

Also noted in the companion but not fixing here since it's a separate, narrower issue: §9 Q5's rung-collision note doesn't fully account for the situation — `reference-app-sdk-packaging.md`'s `ref-08-pnm-core-hermes` collides with the parent plan's own existing `ref-08-rn-hello`, not just with `pnm_cnm_subtask.md`'s `ref-10-pnm-core-hermes`. Worth a follow-up rename in that document.

Happy to talk through the wording changes if you'd rather phrase it differently — the substance is what I care about, not these specific sentences.

PR #29, opened 2026-09-04
feat: require signed and DCO-signed-off commits

bmiller59:
## Summary
- `.husky/pre-commit` now also runs `scripts/check-commit-signing.sh`, which fails fast if `commit.gpgsign` isn't enabled locally. `.husky/commit-msg` already enforces `Signed-off-by:` via commitlint.
- Add `.github/workflows/commit-checks.yml`, a PR check that walks every commit in the pull request via the GitHub API and fails if any (non-Dependabot) commit is missing `Signed-off-by:` or isn't GitHub-verified as cryptographically signed. Dependabot commits are exempted from the sign-off check (it can't add one) but still must be verified-signed.
- `scripts/ensure-bifold-ready.js` now also wires up the bifold submodule's own hooks (`git config core.hooksPath .githooks`) on every root `yarn install`, once berkmancenter/keyring-bifold#42 lands.
- Document the requirement in `CONTRIBUTING.md` and `CLAUDE.md`.

Companion PR for the submodule: berkmancenter/keyring-bifold#42 (same approach: `.githooks/`, `commit-checks.yaml`, CONTRIBUTING).

A git hook can't itself verify a commit's signature — git signs the commit object in `commit-tree`, which runs after every hook fires — so the pre-commit hook only checks that signing is *configured*. The actual signature is verified server-side: by CI on every PR now, and (once in-flight PRs on `main` are clear) by branch protection requiring signed commits, which rejects an unsigned commit outright. That branch-protection change is **not** included here since flipping it live today would immediately block every currently-open PR on `main` (several of which predate this change) until each is rebased with signed + signed-off commits — see PR discussion for the exact command to run once the queue is clear.

## Test plan
- [x] `yarn test` (full suite, via pre-push hook) — passed
- [x] `yarn pre-commit-lint` / `yarn typecheck` (via pre-commit hook) — passed
- [x] Committed this change itself with the new pre-commit signing check active, signed and signed off
- [ ] CI run on this PR (commit-checks.yml) — will self-validate once opened

### DIFF
--- a/.github/workflows/commit-checks.yml
+++ b/.github/workflows/commit-checks.yml
@@ -1,0 +1,57 @@
+name: Commit Requirements
+
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened, edited]
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  dco-and-signature:
+    runs-on: ubuntu-22.04
+    name: DCO sign-off and commit signature
+    steps:
+      - name: Check every commit in this PR is signed off and cryptographically signed
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+          fail=0
+
+          commits=$(gh api "repos/${REPO}/pulls/${PR_NUMBER}/commits" --paginate --jq '.[] | @base64')
+
+          for row in $commits; do
+            data=$(echo "$row" | base64 -d)
+            sha=$(echo "$data" | jq -r '.sha')
+            author_login=$(echo "$data" | jq -r '.author.login // empty')
+            message=$(echo "$data" | jq -r '.commit.message')
+            verified=$(echo "$data" | jq -r '.commit.verification.verified')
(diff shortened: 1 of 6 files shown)


bmiller59:
### Enabling branch protection once the PR queue is clear

Not enabled yet — flipping it live today would immediately block every other currently-open PR on `main` until each is rebased with signed + signed-off commits. Once those are merged or rebased, run:

\`\`\`bash
gh api --method PUT repos/berkmancenter/keyring-wallet/branches/main/protection --input - <<'JSON'
{
  "required_status_checks": {
    "strict": false,
    "contexts": ["DCO sign-off and commit signature"]
  },
  "enforce_admins": false,
  "required_pull_request_reviews": null,
  "restrictions": null
}
JSON

gh api --method POST repos/berkmancenter/keyring-wallet/branches/main/protection/required_signatures
\`\`\`

This marks the \`commit-checks.yml\` job (added in this PR) as a required status check and enables GitHub's native "require signed commits" on \`main\`.

Note: any SSH/GPG key used to sign commits must be registered on the committer's GitHub account as a **signing key** specifically (Settings → SSH and GPG keys → New SSH key → Key type: Signing Key) — a key that's only registered for authentication will show as \`unknown_key\` and fail the check.

PR #53, opened 2026-09-15
docs(plans): community vetting subtask — Keyring as a PNM on a local VTI stack

albertoleon7794:
A new subtask of `keyring-on-the-vta-farm.md`, its dated companion, and three pointer edits in the parent (header, §9 Q3 marked answered, review-index row). Docs only.

## What it proposes

Keyring as the **applicant** in the ecosystem's peer identity vetting ceremony — vetter ticket → session with a spoken match code → Vetting Card → Vetting Statement → join submit → membership credential and reciprocal VMC — on a phone, targeting OSS Summit Europe (Prague, 2026-10-01). Upstream's own V0 target for the ceremony is 5 October, driven from a terminal client.

Keyring becomes a **PNM**: a new bifold `vti-client` package wrapping `@openvtc/pnm-core` (0.9.1 already implements the member side of joining; it has no vetting client yet), carried over the Credo DIDComm v2 line.

## Why a local VTI stack first

The newest images a Farm session can select (`vta`/`vtc` `0.24.1-0af2cce7`, 2026-09-09) **predate the vetting implementation**, so a Farm cannot run the ceremony yet. Development and e2e run against a local four-service stack — three VTAs, VTC, mediator, DID hosting — built from VTI `main` (vta-service 0.27.0 / vtc-service 0.11.58 / vta-sdk 0.38, the versions the vetting dry-run runbook is written against) and configured from the Farm's own service templates so it transfers.

**Setup note for anyone running it:** every service must be reachable on a stable public HTTPS hostname (upstream now refuses `did:webvh` on localhost/private hosts), so each developer needs **six reserved ngrok dev domains** of their own (§4).

## Reading guide

| Section | What's there |
|---|---|
| §2 | Positions: pnm-core in bifold, local stack vs local Farm, Farm images, keys (persona keys on the VTA; device admin key software first, hardware target), test counterparties are upstream's own code |
| §3.1–3.4 | The membership and vetting wire, cited to public upstream docs and tests |
| §3.5 | **Conditional design** — upstream's client-side vetting is still being iterated |
| §3.6 | **TSP Rev 3** is a hard cutover (HPKE-Auth removed); the vetting path is designed not to depend on TSP |
| §3.8 | The *Vetting Dry Run* runbook: topology, twelve steps with expected outcomes, assertion table, negative cases — the acceptance script for P2/P6 |
| §4 | The local stack, including the tunnel/domain setup |
| §6 | Phases P0–P8, each with done-when and blocked-on |
| §7 / §7.1 | UX: My Agent tab, one scanner, ceremony vocabulary, full-screen match code, biometric at every signing act, legal-name handling, refusals — and a screen-by-screen table (S1–S13) |
| §8 | e2e runners on Android emulator and iOS simulator, negative cases |
| §9 | Requests to upstream, sent together after P6 |
| `2026-09-15-al.md` | Findings F1–F10 with evidence, and what each supersedes in the parent |

## Feedback most wanted

- The architecture choice in §2.1/§5 (pnm-core inside bifold vs. our own client) and the React Native seam list.
- Whether P2's reference run and the headless vetter (P3) are the right oracle before Keyring work starts.
- The screen flow in §7.1.
- Anything in §3.8 that doesn't match what you've seen the services do.

---

# Added 2026-09-15 (BM) — a second subtask, and this session's review companion

`7c2d394` adds a second subtask under the same parent, plus the dated companion for the review of the one above. **The PR is now two subtasks, not one** (6 files, +854/−2); everything above still describes the community vetting subtask and is unchanged.

## `tsp_rev3_subtask.md` — the TSP Rev 2 → Rev 3 cutover

Opened because §3.6 above records Rev 3 as a hard cutover the vetting path is designed *not* to depend on — which is the right call, and leaves the cutover itself unowned. Parented here because the Prague path forces the question; [`openvtc-integration-plan.md`](../blob/main/docs/plans/openvtc-integration-plan.md) keeps ownership of TSP as a transport.

**The headline is that upstream already answered the design question.** `OpenVTC/vta-browser-plugin` has a branch `feat/tsp-rev2-rev3-dual-handler` @ `78cfbf96` (unreleased `@openvtc/vti-tsp-js` 0.3.0) whose changelog states it: *"This package now packs Rev 3, and a Rev 2 peer cannot read what it sends. There is no negotiation and no fallback. … Reading is dual."* So the plan adopts **pack one revision, read two** unchanged, and keeps only the half the changelog leaves to callers — persisting a peer's observed revision as a diagnostic, never feeding it back into what we pack.

**Two consequences are ours rather than upstream's:**

| | |
|---|---|
| **The mediator demux breaks above 12,285 bytes** | Rev 2's `-E` count covered only the envelope header, so a TSP frame always began `0xF8`/`"-E"`. Rev 3 widens it to cover the ciphertext, so anything larger is framed long — `0xFB`/`"--"`. `tsp-js` fixed its own `isTsp`; the classifier we ride is in another package and was not — `vti-didcomm-js/src/mediator-transport.js:536` routes on `text.startsWith("-E")` and otherwise falls through to the DIDComm unpacker. Measured, not inferred (§3.3, and the companion's appendix is runnable). **The threshold is inside our range** — our attestation fixture is 7,346 bytes and a captured witnessed edge 6,623, so a witnessed exchange carrying both parties' chains crosses it, and it fails *silently* as a DIDComm decode error. |
| **Relationship gating stops being optional** | *"Rev 3 gates application messages on a relationship, so without these a peer enforcing §7.2.2 drops everything a client sends — silently, since a dropped message answers nothing."* This supersedes `pnm_cnm_subtask.md` §6 item 5, which records the `XRFI`/`XRFA`/`XRFD` FSM as "declared-but-dead". |

Also recorded: Rev 3 takes the static X25519 DH off the send path (confirmed at our `hpke.ts:110`), so a sending device needs only a signing key — which means the hardware-custody target belongs here rather than in **§2.4/P4 above**, where it would be measured under a Rev 2 constraint Rev 3 retires.

Phases **R0–R7** each carry acceptance criteria and a blocked-on. No dates: the work is gated on an upstream release, and the Prague path is designed not to wait for it. **R1** turns the demux check into a runnable reference rung in the ladder's existing shape (`run.mjs`, `npm run check`, joining the sweep), with an offline mode that needs no mediator so it stays green in CI, and an opt-in live mode. Five requests to upstream, including publishing 0.3.0 and fixing the transport classifier.

## `2026-09-15-bm.md` — one companion, two parts

Per `docs/plans/CLAUDE.md` (*"A session that produces both a review of the plan and design decisions for a subtask writes one file with two parts"*):

- **Part 1 — review of the community vetting subtask**, findings **R1–R12**: what its evidence base verifies (all four baseline SHAs resolve; F1, F2's conclusion, F4 and F6 all check out), then the gaps — chiefly the DIDComm v2 carriage reversing a standing position in two siblings without a record, no gate for a Prague that precedes upstream's own V0, `requirementsGrace` unhandled, the PNM surface and VTA management scoped by silence, and §3.8's runbook not being vendored. Raised in the review comment on this PR; recorded here because a PR comment is not where the next implementer looks.
- **Part 2 — Rev 3**, findings **T1–T8**, including T1's record of a position I proposed in review and then abandoned (a per-VID send policy) once the upstream branch showed why a dual *packer* can only be a guess.

## Note

The PR title still names only the vetting subtask. Happy to leave it, retitle it, or split the Rev 3 work into its own PR — @albertoleon7794's call.


bmiller59:
## Review — design choices vs. possible gaps

Read the subtask, the companion and the three pointer edits, and checked the evidence base against upstream rather than taking it on trust.

**Updated 2026-09-15:** TSP Rev 3 was excluded from the original comment because we were taking it separately. It has since been taken — as a second subtask on this same branch (`7c2d394`) — so section **E** below now covers it, and the one gap it opens in the vetting subtask is recorded there rather than left implicit.

The citations hold where I could check them. All four baseline SHAs resolve and are same-day: openvtc `9a2d174e`, VTI `53a7cde4`, `dtgwg-trust-tasks-tf` `6e667c1d`, `vta-browser-plugin` `21b0465`. F1 confirms — image commit `0af2cce7` is 09-09, the three vetting commits are 09-11/09-12, so a Farm genuinely cannot run the ceremony. `submit/0.2/spec.md` confirms the applicant as proof signer and the `refer`/`requestMore` split verbatim; `manifest/0.2/spec.md` confirms `vetting`, `branding` and `requirementsDigest`. F4 and F6 check out against this repo. This is a strong document, and everything below is about seams between plans rather than quality.

Structure below: **design choice** = decided with rationale, and we're comfortable with it. **Possible gap** = something a reader six months from now gets wrong without it.

---

### A. Carriage over Credo DIDComm v2

**Design choice — accepted.** DIDComm v2 as the default carriage for every leg, `pnm-core`'s own channels as the fallback where that path doesn't reach, and per-leg selection measured in P4 rather than assumed.

**Possible gap — the position reversal isn't recorded anywhere.** `openvtc-integration-plan.md:101` states *"DIDComm: v1 only (Credo has no v2)… Keyring↔VTA therefore uses OpenVTC's JS libraries beside Credo, **never through it**"*, and `pnm_cnm_subtask.md:977` repeats it: *"It requires DIDComm v2, **which Credo cannot speak** (parent §3)."* §5 routes it **through** Credo. The v2 line is essentially done on another branch, so the code isn't the issue — the record is. As it stands three plans contradict each other and the newest one wins silently, which is exactly what `docs/plans/CLAUDE.md` says a design position adopted must not do.

*Fix:* an F-finding in `2026-09-15-al.md` — *Credo speaks DIDComm v2 on the 0.7 line; evidence: `<branch/PR>`; supersedes `openvtc-integration-plan.md` §3's "Credo has no v2" and `pnm_cnm_subtask.md` §6 item 4's "beside Credo, never through it"* — plus a one-line correction in each sibling pointing at it.

**Possible gap — P0 says `Blocked on: nothing`** while building on "a checkpoint of the Credo 0.7 / DIDComm v2 line", and the document gives no way to find that line. I went looking: nothing named `ref-14`…`ref-19` on `main` or any of 116 remote branches, and the only DIDComm branch here is `feat/trust-tasks-over-didcomm-v1`.

*Fix:* name the branch and the checkpoint SHA in P0, the way §4 pins everything else.

**Possible gap — no phase owns the v2 mediator transport.** "Mediator login, one socket per DID, Pickup 3.0" appears once in §5 and then only as a clause inside P4's done-when (*"mediator session that persists an inbound message before acknowledging it"*), on a three-day phase.

*Fix:* either state it's inherited complete from the v2 line and cite the rung that proves it — §5's *"reference rung on that same line"* is one of the unlocatable ones — or give it its own done-when.

---

### B. How much of the vetting ceremony

**Design choice — accepted.** The applicant half only; the vetter is upstream's Rust code rather than a second copy of Keyring (§2.5); Keyring-as-vetter is later work (F3); four of nine negative cases in P6 with the rest in §8's "later" row. §3.5's conditional design is a model of what the conventions ask for — the oracle is named, and a gap becomes a request rather than a local patch.

**Possible gap — upstream's V0 is four days after Prague, and the consequence is never drawn.** §1 states both dates (V0 5 October, Summit 1 October). P6 lands 09-27 and is `Blocked on: §3.5`, which itself predicts *"at least one further upstream iteration before P6 is final."*

*Fix:* a decision gate saying what ships on 10-01 if P6 is red on 09-27. P5 (plain membership) is the obvious committed deliverable — it needs neither vetting nor, per §2.3, the local stack.

**Possible gap — `requirementsGrace` is unhandled.** `manifest/0.2/spec.md`: *"Because vetting takes days or weeks, an applicant starts under one version of the requirements and may finish under another"*, and a community MUST evaluate under the digest's version only while that version is inside `requirementsGrace`. S7 has a *refresh requirements* action; nothing defines the out-of-grace state. In a ceremony designed to span weeks, that's the state most likely to strand a real user.

*Fix:* a §3.3 sentence and a corresponding S7/S13 state.

**Possible gap — which wire a deployed VTC actually serves.** F2 cites `ceremonies/vtc/member-onboarding/0.1/ceremony.json` @ `6e667c1d` for the 0.2 versions, but that file still wires `manifest/0.1`, `submit/0.1` and `submit-receipt/0.1`. The 0.2 specs exist separately at `specs/vtc/join-requests/{manifest,submit}/0.2/`, so the conclusion is right and the citation points at the wrong file. The lag is itself the interesting part: the ceremony definition trails the task specs.

*Fix:* re-cite to the spec paths, and raise the ceremony/spec version lag in §9 now rather than leaving it for P2 to rediscover. (§3.1's table also omits the ceremony's `decide/0.1` and `solicit-vmc/0.1` while saying it is "From [[ONBOARDING-CEREMONY]]".)

**Possible gap — no vetter discovery.** `manifest/0.2/spec.md` points an applicant who knows no vetter at `vtc/vetting/vetters/list`, and step 06 mentions publishing a vetter profile — but S7 only offers scan/paste a ticket.

*Fix:* an S7 affordance, or a sentence saying ticket-only is deliberate for V0.

---

### C. The PNM surface

**Design choice — accepted.** Wrap `@openvtc/pnm-core` behind a DI token rather than reimplement (§2.1), with the architecture owned by `pnm_cnm_subtask.md`, and 0.9.0 avoided for its partial tarball. The RN seam list is hedged correctly — `redirect: "manual"` is *"to be measured before it is relied on"*, which matters precisely because the egress guard's no-redirect guarantee rests on it.

**Possible gap — the plan never says what it isn't building.** `pnm_cnm_subtask.md` §6 P3 defines the PNM surface as credential vault, device management (`device/{register,heartbeat,list,disable,wipe}`), session revocation (`auth/revoke-session`), approvals, VTA seeds, backup/restore and memory. None are in scope here and nothing says so — the only "later work" in the entire document is push notification. A reader arriving from the sibling plan will assume they ride along.

*Fix:* one sentence in §2.1 or §5 listing what P4/P5 deliberately omit, and which of it, if any, Prague needs.

**Minor:** F6's "22 of 25 published subpaths" — npm reports 24 subpaths in 0.9.1's exports map. The ratio is the argument, so it's worth re-deriving.

---

### D. Connecting to and managing a VTA

**Design choice — accepted.** Keys held by the VTA as upstream designs them (§2.4); software Ed25519 first with the hardware target chosen by P4 measurement; biometric at every signing act; the legal name written to a VTA face marked sensitive and not duplicated in Keyring. §9's requests 2 and 3 correctly name what's unanswered, and P8 is honestly blocked on the Farm operator.

**Possible gap — "manage" is unscoped.** Across S1–S13 the single write to the VTA is S8's legal-name face; everything else is status and lists. That's a reasonable V0, and the document never says it is one.

*Fix:* a §7 sentence — My Agent is connect-and-observe for P4–P6, and VTA lifecycle management belongs to `pnm_cnm_subtask.md`.

**Possible gap — no recovery path.** Phone lost, admin key rotated, re-enrolment on a new device. §9 request 3 asks upstream whether a VTA may carry one grant per device, but no screen would use the answer and no device-grant revocation exists.

*Fix:* name it out of scope with that request as its blocker, or add it to P4.

**Possible gap — the legal-name face has no exit.** §7 handles collection well and says nothing about an abandoned application, deletion, or what the vetter retains after attesting. Prague is in the EU.

*Fix:* a retention sentence in §7 and, if upstream has no deletion task, a §9 request.

---

### E. TSP Rev 3

**Design choice — accepted, and the best-insulated decision in the document.** §3.6 puts Rev 3 out of scope with an owner named — *"Its migration belongs to `openvtc-integration-plan.md`, triggered by upstream's Rev 3 release, and is not scheduled here"* — and the insulation is structural rather than hopeful: *"Every leg has a DIDComm v2 route; TSP legs are optional until Rev 3 settles."* Fixtures record their revision. F5 carries the evidence and the superseded position. Nothing about the vetting path can be hurt by the cutover, which is exactly right.

Researching what the cutover costs turned up three things that bear on **this** PR.

**Upstream has already settled the design question.** `OpenVTC/vta-browser-plugin` carries `feat/tsp-rev2-rev3-dual-handler` @ `78cfbf96` — unreleased `@openvtc/vti-tsp-js` 0.3.0 — whose changelog states it: *"This package now packs Rev 3, and a Rev 2 peer cannot read what it sends. There is no negotiation and no fallback. … Reading is dual."* So it is neither automatic on the wire nor a toggle: **pack one revision, read two**, with the revision readable in nine bytes without keys. Worth knowing here because it means no revision-selection mechanism is ours to design.

**Possible gap in this subtask — §2.4's hardware-custody sequencing.** §3.6 correctly notes Rev 3 *"removes the static X25519 key a device could not hold in hardware from the send path."* I confirmed that against our own code: `hpke.ts:110` (`authEncap`) is the only place the send path consults `senderKeyAgreement.agree()`, and under HPKE-Base the sender does no static DH at all. The consequence §3.6 doesn't draw is that **P4 picks a hardware target by measurement among three Rev 2-shaped options**, and Rev 3 retires the constraint that shapes them.

*Fix:* either say P4's measurement is deliberately Rev 2-only and will be revisited, or move the choice to the Rev 3 subtask, which is where it now lives.

**Not a gap here, but it lands near this plan's payloads.** Rev 2's `-E` count covered only the envelope header, so a TSP frame always began `0xF8` / `"-E"`. Rev 3 widens it to cover the ciphertext, so anything larger is framed long — `0xFB` / `"--"`. `tsp-js` fixed its own `isTsp`; the classifier we ride is in a different package and was not — `vti-didcomm-js/src/mediator-transport.js:536` routes on `text.startsWith("-E")` and otherwise falls through to the DIDComm unpacker. Measured, not inferred; the boundary is **12,285 bytes**, our attestation fixture is 7,346 and a captured witnessed edge 6,623, so a witnessed exchange carrying both parties' chains crosses it — and fails silently, as a DIDComm decode error. Filed upstream as a request, and the check becomes a runnable rung rather than a snippet.

Also: Rev 3 *"gates application messages on a relationship"*, so a peer enforcing §7.2.2 silently drops everything an unaware client sends. That supersedes `pnm_cnm_subtask.md` §6 item 5's "declared-but-dead" reading of the `XRFI`/`XRFA`/`XRFD` FSM — another instance of the seam problem in the closing note below.

---

### Cross-cutting

- **§3.8 rests on a source nobody can open.** [[DRY-RUN]] is *"shared with the team as a PDF. Not vendored here"*, yet its *Expect* rows are the done-when for P2 and P6, and individual assertions carry no citation (*"five bad codes from one sender in an hour drops that sender silently for the hour"*). `docs/plans/CLAUDE.md`: *"Every normative constraint carries a citation… Where a plan depends on more than a handful of external requirements, extract them into a constraints companion."* **Fix:** vendor the steps, assertion table and negative cases into a constraints companion beside the plan.
- **`external/` cannot verify a single citation today.** The pinned clones sit at `187ad9cd` / `89d70c4` / `7e0d755` (mid-August) and none of them resolve the four baseline SHAs; I checked via `gh api` instead. §4 advances pins only in P0. **Fix:** advance the pins in this PR, or state in the Baseline line that they run ahead of `external/` and how to check them.
- **Rung numbering (F8) closes a smaller collision than the one already on `main`.** `tsp-reference/README.md:75-78` documents `ref-10-credo-askar-adapter`, `ref-11-vidresolver-port`, `ref-12-direct-ts-port` and `ref-13-jcs-canonicalizer-corpus` — a four-way collision with `pnm_cnm_subtask.md` §6's reserved `ref-10`…`ref-13`. Starting at `ref-20` dodges it; it doesn't close it. **Fix:** let `tsp-reference/README.md` own the allocation and have plans claim from it.
- **Where do §8's runners run?** Six reserved ngrok domains *per developer* means these five e2e scripts cannot run in CI as designed, and the parent keeps `ref-05-local-vta` as the CI environment precisely because the Farm can't be a test dependency. **Fix:** say they're attended-only, or name the CI story.
- **Eight broken links.** The header and §2.1/§2.2/§3.6/§5 wrap whole markdown links in backticks (`` `[name](path)` ``), so they render as literal code rather than links — 8 in this file, 0 in the parent, the companion or `pnm_cnm_subtask.md`.

---

Nothing here blocks the direction. If only two things change before merge, I'd still pick the **Prague fallback gate** (B) and the **F-finding recording the Credo carriage reversal** (A): the first changes what we commit to on 10-01, and the second is the one a future implementer cannot reconstruct.

Reading across A–E, the gaps cluster where a plan **inherits** rather than decides — the DIDComm v2 line from a sibling, the PNM surface from `pnm_cnm_subtask.md`, VTA management from the parent's F2, the Rev 3 constraint from §3.6's deferral. Each of those is sound read alone, and the seam between them is where nobody is looking. E is now owned by its own subtask on this branch; A–D are still open above.

All findings are recorded in `2026-09-15-bm.md` Part 1 as R1–R12, so they survive this thread.

bmiller59:
## Where this work could run in parallel

Read §6's dependency lines as a graph rather than as the column of dates they're written in. Offered as a proposal, not an edit — two of the items below are calls only @albertoleon7794 can make.

Short version: **the two biggest wins are decouplings, not extra people**, and one of them can start today rather than on 09-22.

### The graph is already more parallel than the plan reads

The targets run 09-16 → 09-29 in a single column, which reads as a chain. The `Blocked on:` lines actually describe two chains that fork after P1 and rejoin at P6:

```
P0 ─ (independent)
P1 ─┬─ P2 → P3 ─────────┐
    │  (Rust / CLI / ops)├─ P6 → P7 → P8
    └─ P4 → P5 ─────────┘
       (React Native)
```

That fork has a natural skills split — P2/P3 are `cnm`, REST and upstream's own Rust; P4/P5 are RN, Credo and Hermes. Little overlap, different people. Since the plan doesn't say so, it is likely to be read and staffed as a sequence.

### One thing to resolve first: P4's stated blocker doesn't match its gate

P4 reads `Blocked on: P1`, but its done-when requires *"byte-equality of Keyring-built documents against **P2's fixtures**"*. So as written P4 cannot complete without P2, and the two chains above are not actually independent.

Which way this is fixed decides how much parallelism is really available:

- move the byte-equality assertion to **P5's** gate → the fork is genuine, and the RN chain can run well ahead of the ceremony chain; or
- keep it at P4 → P4 is blocked on `P1, P2`, and the plan should say so, because the current line understates it.

Your call — flagging rather than editing.

### Decoupling 1 — split P4's platform half out; it needs no infrastructure

P4 bundles two unrelated kinds of work behind `Blocked on: P1`:

| Needs only a device | Needs the stack |
|---|---|
| the eight RN seams on Hermes — `node:fs` shim, `crypto.subtle`, `getRandomValues`/`randomUUID`, `AbortSignal.timeout`, `structuredClone`, the KV adapter, and especially whether RN's `fetch` honours `redirect: "manual"` | enrolment by QR, authenticated session, agent health, persona list |
| `did:webvh` on Hermes against **fixture** DIDs, plus the tampered-log case | the same against the stack's real DIDs |
| the §2.4 hardware-custody measurement — can StrongBox hold Ed25519, does the Secure Enclave yield a usable P-256 `did:key` | "and does the VTA accept it" |

The left column is about half of P4's risk and the half most likely to produce an unpleasant surprise — `redirect: "manual"` in particular, since §5 already notes the egress guard's no-redirect guarantee depends on it and it is *"to be measured before it is relied on"*. None of it needs a VTA, a mediator or a tunnel. **It can start now; today it is scheduled to start 09-22.**

### Decoupling 2 — stage P1 rather than treating the stack as one deliverable

The full topology is three VTAs + VTC + mediator + DID hosting. But **P5 needs one VTA**; only the ceremony needs three.

- **P1a** — single VTA + VTC + mediator + DID hosting → unblocks P4/P5
- **P1b** — the second and third VTAs and the two-profile topology of §3.8 → unblocks P2/P3/P6

Today everything Keyring-side waits on the complete six-service stack.

### The ceiling on parallelism, which nothing can raise

§4 is explicit: six reserved ngrok domains **per developer**, because *"a reserved domain terminates on one agent at a time and DIDs are bound to their hostname — two developers cannot share one stack's domains."*

So live-stack work does not parallelize cheaply — each extra person needs their own six domains, their own build and their own provisioning. That points the whole strategy one way: **maximise what runs against fixtures or a device, and minimise how many people need a live stack.** Which makes two more chunks attractive, neither currently named as separable:

- **The 13 screens (S1–S13), three locales, the accessibility bar.** These consume data shapes, not services. P2's frozen fixtures are already the stated contract for byte-equality, so they are equally a contract for the UI — build the screens against fixtures and wire them last.
- **The e2e harness** — `e2e/lib/vti-stack.js`, deep-link QR injection, runner scaffolding — against stubs, before the flows exist.

### What I would not parallelize

P6 genuinely needs P3 **and** P5; that join is real. And §3.5's uncertainty — *"expect at least one further upstream iteration before P6 is final"* — does not shrink with more people on it. P2/P3 are discovery, where a second pair of hands mostly adds coordination. With sixteen days to Prague, extra lanes have a real cost; the decouplings above are cheaper than staffing.

### The shape I would propose

Five tracks, only two of which ever need a live stack:

| # | Track | Needs | Feeds |
|---|---|---|---|
| 1 | **Platform** — seams, Hermes `did:webvh`, hardware custody | a device, nothing else. **Starts now** | P4 |
| 2 | **Stack** — P1a, then P1b | ngrok domains, ops | everything |
| 3 | **Ceremony oracle** — P2 → P3 | P1b | P6, and the fixtures |
| 4 | **Client + UI** — plumbing on P1a; screens on P2 fixtures | P1a, fixtures | P5, P6 |
| 5 | **Harness** — e2e scaffolding, plus TSP Rev 3's R1 | nothing | P5–P7 |

Main risk is fixture drift: if P2's output moves, track 4 reworks. The byte-equality contract bounds that, and it argues for P2 freezing fixtures early even where the ceremony is incomplete.

### What the Rev 3 subtask contributes here

Almost nothing, by design — R2–R7 are a real chain behind a release that does not exist. Two exceptions: **R0** (minutes, just recording positions) and **R1**, the demux rung, which is blocked on nothing, needs no Rev 3 crypto, no mediator and no mobile. It is a self-contained day for whoever is not on the Prague critical path, which is why it sits in track 5.

---

If this is worth adopting, it changes phase scope and acceptance criteria — substantive under `docs/plans/CLAUDE.md`, so it wants a companion entry recording why rather than a silent §6 rewrite. Happy to draft that plus the §6 revision, or to leave §6 alone and keep this as thread context. The P4 gate question above is the one thing I would like an answer on either way, since the current line understates the blocker.

albertoleon7794:
Agreed on the Rev 3 subtask — happy for `tsp_rev3_subtask.md` and `2026-09-15-bm.md` to ride on this branch and land with it.

I re-checked the load-bearing parts and they hold:

- `@openvtc/vti-tsp-js` 0.3.0 exists only on `feat/tsp-rev2-rev3-dual-handler` (78cfbf9); npm is still 0.1.0/0.2.0, both Rev 2.
- `revision.ts` reads the revision keyless in nine bytes (`f8 00 01` Rev 2, `f8 00 02` Rev 3).
- The demux finding reproduces: `vti-didcomm-js` `mediator-transport.js:536` classifies with `text.startsWith("-E")`, and Rev 3's long-form count puts anything past 4095 quadlets (12,285 bytes) on `--`/`0xFB`. Our attestation fixture is ~7.3 KB on its own, so a witnessed exchange with both chains crosses it. Silent misroute, and it would read as a message-format bug — good catch, and the right thing to pin as a rung.
- On our side: `hpke.ts` has `authEncap` (line 102, called at 137), and we're on the 0.1.0 Hermes patch as you describe.
- Your correction to `pnm_cnm_subtask.md` §6 item 5 is right: the relationship markers stop being dead under Rev 3.

Two small things for R1/R0 rather than blockers:

1. The rung name `ref-2x-rev3-demux` will collide with the `ref-20`/`ref-21` this branch already claims. Let's settle allocation in `tsp-reference/README.md` first (your cross-cutting point on rung numbering) and number it from there.
2. Worth a line in the subtask header that `openvtc-integration-plan.md` keeps ownership of TSP and this is parented here only because Prague forced the question — you say it in the Siblings line, but it's the kind of thing a later reader checks in the header.

Also agreed that R1 and R0 are the only parts runnable now, and that R1 needs no release, no mediator and no device.

Replying separately on the parallel-work proposal — I want to work through the P4 gate question properly before answering it.

albertoleon7794:
On the parallel-work proposal — agreed, and thanks for reading the dependency lines as a graph. Adopting all three decouplings:

1. **Split P4.** The phone-only half — the eight RN seams, `did:webvh` on Hermes against fixture DIDs, the hardware-custody measurement — needs no infrastructure and starts now rather than 09-22. The rest of P4 (enrolment, session, health, persona list) stays on the stack.
2. **Stage P1** into P1a (one VTA + VTC + mediator + DID hosting) and P1b (the second and third VTAs, the two-profile topology). P1a is what the app lane needs.
3. **Build the screens and the e2e scaffolding against fixtures/stubs**, wired to services last.

**On the P4 gate question — you're right that the current line understates it, and I'd fix it by moving the assertion rather than adding a blocker.** Byte-equality belongs where each document is produced: enrolment and auth documents in the client half of P4, join documents at P5. So P4 keeps `Blocked on: P1a` honestly, and the fork is real.

Suggested split so neither lane waits on the other:

- **Me:** platform half of P4, P1a for my own stack, client + UI.
- **You:** ceremony oracle (P2 → P3), the harness, and R1.

The one genuine handoff is the headless vetter at P6. If it slips I can drive the vetter side manually from the terminal on my own stack, since P2 is a manual run anyway.

Two asks:

- **Freeze P2's fixtures early**, even where the ceremony is incomplete — that's the contract the UI and the byte-equality checks build against, and it bounds the drift risk you named.
- **Please do draft the companion and the §6 revision**, since it changes phase scope and acceptance criteria. Include the mediator position while you're in there: the VTA leg has to use an Affinidi-style mediator (ours locally, whatever the VTA advertises on the Farm) because upstream clients authenticate with the ATM login and a Pickup 3.0 socket that our Credo mediator does not implement — `ref-18` measured that, and `ref-06x` measured the VTI mediator refusing v1 with a 404. Our Credo mediator stays on the wallet-to-wallet legs.

Also agreed on the ceiling: six reserved domains per developer means we each run our own stack, which is fine — it keeps the lanes independent.

albertoleon7794:
Plan updated on the branch (`2bc6baf`, on top of your Rev 3 commit): the DIDComm v2 line is now the named branch point rather than an unnamed local line.

- **keyring-wallet#54** (`feat/credo-0.7`, head `d359e2ba9`) — the credo-ts `0.7.1-pr-2704` hop, the v2 developer setting with its own mediation, the v2 / TSP-over-v2 e2e suites including the two-iOS-device run, and `didcomm_v2_subtask.md`.
- **keyring-bifold#54** (`feat/credo-0.7`, head `5a90b9f36`) — `DidCommV2Carriage`, `v2Routing`, `TrustTaskEnvelopeV2Message`, TSP over a v2 connection, and the mediator and witness serving v2 beside v1.

P0 now branches from those two SHAs and rebases onto whatever they merge as. The assumption is that they land as they stand; anything you find gets folded in as a fix rather than waited on.

This closes most of your first finding: wallet#54 already rewrites `openvtc-integration-plan.md` §3 and its stage-3 row away from *"Credo has no v2 … beside Credo, never through it"*, with the device evidence. The one line still stale is `pnm_cnm_subtask.md` §6 item 4 (*"which Credo cannot speak"*, line 977) — noted in the companion as F11 for whoever touches that plan next.

Two things those PRs deliberately do **not** carry, now stated in the plan so nobody hunts for them:

- the reference rungs `ref-15…19`, which stay local;
- the VTI-mediator transport (mediator login, one socket per DID, Pickup 3.0) — proven in Node only, phase V3 of the v2 subtask. That is why P4 here owns building it for the app rather than inheriting it, which also answers your "no phase owns the v2 mediator transport" point.

bmiller59:
## The A/B decision has a written argument upstream, and it argues for B

Pushed as `e9cc39d` — §2.6, §3.9 and companion `2026-09-16-bm.md`. Flagging the part that bears on F14, since that's the open decision rather than something I should settle in a plan edit.

ic3.software published *The Agent Is the State: UIs That Keep Nothing* (`ic3.software/blog/the-agent-is-the-state`), describing the architecture of their own clients — the PWA and extension that publish `pnm-core`. The VTA holds keys, personas, policies and application state; a UI sends Trust Task documents and renders the replies; authority "comes from the agent, via a who-am-I task, and is never inferred from anything stored locally"; and the instruction to a client author is "do not build a second copy of anything the agent already is."

**That is the written case for option B**, from the people who built the stack. Worth having in front of the decision for two reasons: it tells us what the ecosystem will keep building toward, and it is explicit about the price. Their stated costs are "no cache to invalidate because there is no cache" and "no optimistic rendering and no offline mode, because there is nothing to render from when the agent is unreachable." Acceptable in a desktop browser; a different proposition on a phone at a conference, which is the §2.4 trade in one sentence.

Two measurements that came out of reading it against this baseline:

1. **The disclosure preview under B is already a task.** `vta-sdk` at `53a7cde4` registers `persona/disclosure/preview/1.0` beside `persona/disclosure/present/1.0`, and `pnm-core` 0.9.1 implements both (`dist/persona/disclosure.js`). So §3.3's "disclosure preview, then presentation" is two calls, and S10 has a named source instead of a screen that re-renders the face. Under A the same discipline survives without a task behind it: one function builds the card, the screen shows its output.
2. **`vta/app-state/*` exists and is the mechanism B would use** for progress that must outlive an install — versioned per `(contextId, namespace, key)`, `expectedVersion` preconditions, RFC 7386 merge-patch, `list` with `sinceVersion` as a tombstoned change feed, and a failed precondition that returns the current version *and* value so a conflict resolves without a re-read. Upstream records why it isn't `vta/memory/*`: "forget everything" has to stay safe to ask an agent. Given §2.1, adopting it under B is a port onto our transport, sized by six operations — not a free surface.

**What binds either way**, and what §2.6 now states as architecture-neutral: the community owns membership, the published requirements and the verdict under both options, so none of those is rendered from a last-known value. And Keyring keeps durable local state for what it owns — its credentials (the in-person VRC path has no VTA in it under either option), an identity that survives a restart, inbound documents persisted before the ack, and the in-flight task-id ledger, without which §2.2's retry invariant can't be met.

That last one is where F14 is the interesting evidence. The doctrine's weak point on mobile isn't philosophical — it's that "the UI keeps nothing" and "the app forgot its own `did:peer:2`" are the same sentence read two ways. Both credentials sat at the mediator for a DID the app no longer remembered, which is why P5's first rung is now one identity across restarts. So whichever way §2.4 goes, the state Keyring owns has to be durable; B narrows what that set contains, it doesn't empty it.

Not trying to pre-empt the call. If it helps: the evidence leans B for interop — it's where the ecosystem's specs, tasks and client code already are, and A leaves us implementing the ceremony's signing side ourselves with no upstream counterpart. What would flip it is Prague. B needs the enrolment step (§9 request 2) from someone else's roadmap, and A is already working on a device. "A for 1 October, B as the target" is a coherent answer, but only if it's chosen deliberately, because the Vetting Card is a signed object and the choice decides which side signs it.

§2.6 and §3.9 are written to hold under either option, with the B-only parts marked as such, so nothing in them needs rewriting once the decision lands. `2026-09-16-bm.md` F8 records which of my own premises the last two days of execution moved, and F5 records a correction worth repeating here: both tasks above are absent from `external/`'s VTI pin (`187ad9cd`, 08-17) and present at this subtask's baseline. I concluded upstream didn't have them before checking the pin's date. An absence in a four-week-old clone is evidence about the clone.

PR #54, opened 2026-09-15
feat: DIDComm v2 and TSP over v2 — app wiring, e2e on two physical devices, plan

albertoleon7794:
The wallet half of the DIDComm v2 work: the developer setting and its second mediation, the e2e suites (including the first runner that drives two physical iOS devices), the plan documents, and the submodule pointer.

Depends on berkmancenter/keyring-bifold#54, which carries the carriage, the servers and the attestation fix. **Merge that one first** — this branch's submodule pointer references its head.

## Read it in five commits

| Commit | What it does |
|---|---|
| `chore(deps)` | Resolutions to the `0.7.1-pr-2704` snapshot, three yarn patches, the native bumps 0.7 drags along |
| `feat(app)` | "Enable DIDComm v2" beside the TSP setting; Coordinate Mediation 2.0 against `MEDIATOR_V2_URL` as a second mediation record |
| `test(e2e)` | Six new simulator suites, the two-iOS-device runner, and the harness fixes the devices forced |
| `docs(plans)` | `didcomm_v2_subtask.md`, the dated companion, and the parent plan's stage table |
| `chore` | bifold → `5a90b9f` |

## Start with the plan

`docs/plans/openvtc-integration-plan/didcomm_v2_subtask.md` is the argument: why an unmerged PR is the dependency, the phase ladder (V0 feasibility → V1 the hop → V2 the carriage → V2T TSP over it → V3 the VTI leg), and the acceptance criteria each phase was held to. The dated companion `2026-09-13-al.md` is what actually happened, corrections included.

## Nothing changes unless a developer turns it on

The agent is constructed with `didcommVersions: ['v1','v2']` only when the setting is on, and v2 mediation is a second record beside the v1 one. With the setting off, or `MEDIATOR_V2_URL` blank, the wallet behaves exactly as it does today — and `yarn e2e:vrc:witnessed:ios-devices` (v1) is green on the same build to show it.

One rough edge: both developer toggles relaunch the app, because the agent fixes its DIDComm versions at construction, and nothing tells the user that. Fine on a developer screen; it would need a notice anywhere else.

## The e2e suites are stricter than before

Badge assertions used to log what they found. They now fail the run when a shield is missing, wherever both sides are real hardware. That change immediately caught an iOS App Attest bug the logging-only version had been passing over for weeks — the fix is in the bifold PR, and the log trace is in the companion.

Every log-marker assertion also used to skip iOS devices silently, so an iPhone + iPad run would have passed while checking nothing. Each device's log is now captured and read.

## What is proven

- `yarn lint`, `yarn typecheck`, `yarn test` (118) green; bifold's suites green in its PR.
- Simulators: v1, v2, TSP-over-v2, plain and witnessed.
- Physical devices (iPhone 16 + iPad A16, 2026-09-15): `…:ios-devices` (v1), `…:didcomm-v2:tsp`, and `…:didcomm-v2:tsp:locality` — all green, with Secure Exchange + Verified on both contacts and In-Person on the locality run.

## What is not proven

- **B8, the mixed fleet**: a 0.7 wallet against a 0.6.3 build over v1. Not run, and it is the gap that matters for users who have not updated.
- Android devices, and the gradle lockfile for the native bumps.
- QR scanning of a v2 invitation: the runners paste, so only the paste/deep-link path is measured. A 1145-character `_oob` invitation was accepted on both devices.


bmiller59 (approved):
Approving — code and tests look solid. Verified end-to-end on real hardware this session: added and passed `e2e:vrc:devices:android-only:didcomm-v2`, `e2e:vrc:witnessed:android-only:didcomm-v2`, and `e2e:vrc:witnessed:locality:android-only:didcomm-v2` on two physical Android phones (hardware attestation, DIDComm v2, witness, and real BLE co-presence all confirmed). Six real e2e-harness bugs found and fixed along the way, documented in `docs/E2E_REAL_DEVICE_FLAKINESS.md`.

**Blocking before merge**: the "DCO sign-off and commit signature" check is failing on two pre-existing commits (not from this session's work):
- `5746814` "chore: bifold → f2587cb39..."
- `0289413` "style(app): drop unnecessary semicolons in useBCAgentSetup.test.ts"

Both are missing the `Signed-off-by` trailer. @albertoleon7794 — could you add sign-off to these (rebase + amend, since they're not the branch tip) before merge? Everything pushed this session already carries a correct trailer.

PR #55, opened 2026-09-17
docs(plans): migrate the witness credential from the WD02 type to the VSC

bmiller59:
> **Base branch:** this targets `feat/credo-0.7`, not `main` — the branch was forked from it, so it stacks on #54. Merge that first.

## What this is

A plan, not code. Three documents plus an index row; no source file changes.

The DTG Core Credentials spec ([`dtgwg-cred-spec`](https://github.com/trustoverip/dtgwg-cred-spec), **Working Draft 0.4.0**, read at `994a3d63`) collapsed `EndorsementCredential` and `WitnessCredential` into a single `StatementCredential` whose meaning is carried by an absolute-IRI `predicate`. **The name VWC survives** as a predicate profile; the type string does not.

| File | |
|---|---|
| `docs/plans/vsc-migration-plan.md` | The living plan |
| `docs/plans/vsc-migration-plan/cred-spec-constraints.md` | C1–C16 — every normative requirement quoted and cited at a stated commit |
| `docs/plans/vsc-migration-plan/2026-09-16-bm.md` | F1–F15 — measurements, positions adopted, what they supersede |

## The shape of it

Eight wire deltas (§3). The surface is **16 non-test source files**, four of which touch bytes — not the ~2,200 "VWC" occurrences a grep suggests.

Three things the spec now depends on **do not exist yet**: the `dtgwg-predicate-vocab` registry (confirmed absent from the `trustoverip` org), the namespace IRI ([#48](https://github.com/trustoverip/dtgwg-cred-spec/issues/48)), and the DTG `@context`. The bridge mirrors the registry format [#52](https://github.com/trustoverip/dtgwg-cred-spec/issues/52) already specifies, behind one exported constant, split into a **deletable mirror** and a **permanent consumer**. Because the accept-list is verifier *configuration* per *Predicate Handling*, accepting two IRIs while issuing one is conforming — which turns #48 into a rolling cutover rather than a flag day.

Reference rungs `ref-21`…`ref-23` come before any product change; phases V0–V5 each carry acceptance criteria.

## Three findings worth a reviewer's attention

- **Every shipped proof suite is RDF-canonicalized** — `Ed25519Signature2018` and `eddsa-rdfc-2022`, *not* the `eddsa-jcs-2022` the parent plans discuss. An untermed member is not merely unsigned, it is absent from the signed graph. This is the plan's stated highest risk and it drives the V2/V4 split (§3.7).
- **`verifiable-trust-infrastructure` ships a default `personhood.rego` matching `"WitnessCredential" in cred.type`** — one shipped surface, not three; the other occurrences are below each file's `#[cfg(test)]` boundary, and `join.rego` is unaffected. V4 gains a shape switch. **True at our pin only** — that clone's fetch is currently failing, and repairing it is a precondition of V4 (§7.1).
- **The digest change is three breaks, not one.** `ref-07` records it as "encoding-only"; that was true at the pin and is now wrong — WD 0.4.0 hashes the referenced credential *excluding its top-level `proof`* (§3.1).

## Decided in the drafting session

Issue under the spec's placeholder namespace; switch `DTG_CONTEXT_URL` to the spec's IRI; extension members (`locality*`, `hardwareAttestationIncluded`) sit beside `witnessContext` under a Keyring namespace; VDC/VAC/VIC permanently out of scope. Plan §9 splits *Settled* from *Open*.

## Open, for review

1. **When the shape switch flips to `vsc` by default** — the only item on the critical path, and it needs the `verifiable-trust-infrastructure` fetch repaired first.
2. Whether Prague has any witness-shape dependency beyond `join.rego` — I verified the default join policy, not what a Farm-hosted community's custom policy might do.
3. `ref-07`'s capture tool (`captureEdge.local-al.test.ts`) was never committed. Does anyone still have it? Cheaper to recover than rebuild.

## Notes

The `external/` pin is **deliberately not advanced** — everything is read via `git show`. The advance belongs at the start of V0, as one pin with one reason, not bundled with the five other tripwires the same sync reported.


albertoleon7794:
Reviewed, and the design is adopted as written — nothing below reverses anything in §9's *Settled* table. I pushed a companion to the branch rather than writing it all out here: [`docs/plans/vsc-migration-plan/2026-09-17-al.md`](https://github.com/berkmancenter/keyring-wallet/blob/plan/vsc-migration/docs/plans/vsc-migration-plan/2026-09-17-al.md) (`39521e0`), with the pointer added to the plan header in the same commit. Findings are `A1…A12` so they don't collide with your `F1…F15`.

**The one that changes §5.** Five rungs on this subject already exist, and none of them is on `main` — which is why the reserved `ref-07…09` range looks empty:

| Rung | Where | Date |
|---|---|---|
| `ref-07b-statement-shape`, `07c-predicate-coverage`, `07d-vocabulary-trust-path`, `07e-crossimpl-vwc` | branch `feat/ref-07-crossimpl` @ `b8dd46c` — **not pushed** | 09-10 |
| `ref-07f-predicate-registry` | worktree, **uncommitted** | 09-14 |
| `ref-07g-outcome-evidence-pairing` | pushed | 09-15 |

`ref-07f` builds #52's registry and drives predicate handling from it against Keyring's real compiled `credentialTypes.ts`, with zero requests at verification — that is most of `ref-22`. `ref-07b` Act 1 and `ref-07e` Act 4 cover much of `ref-21`. Pushing those is on me, and A1's table is unverifiable until I do; the rolling-cutover four-state check is genuinely new.

**Where the evidence is worse than the plan says.** §3.7's risk is measured in `ref-07c`, and the zero-quad case is the *safe* one — safe mode catches it. An unresolved CURIE is not dropped: `example:observedDocument` is a valid absolute IRI whose scheme is `example`, so it is signed verbatim over an IRI its author did not mean, and expands differently for a verifier holding a context that maps `example:`. So V2's guard has to compare expanded IRIs, not count quads. Good news from the same rung: `digestMultibase` is already `security#digestMultibase` and `@protected` in `credentials/v2`, so V2 needs one term fewer and must not redefine it.

**Four gaps in the ladder:**

1. **Credo 0.7.** `ref-21` says outright it proves nothing about Credo, so the first real sign-and-verify of a VSC is V4's e2e — after issuance flips. This branch sits on `feat/credo-0.7`, where the witness ceremony already depends on `verifyCredentialSubjectAuthentication: false` (D6 adds a subject-binding check next to one we switched off), the snapshot carries main's `w3c-di` restructure, and JCS canonicalization is pinned around a broken publish. Proposed `ref-07h`: a VSC through Credo with our patches, under both suites we ship. Related: `ref-07e` Act 1 found our rung-local `eddsa-jcs-2022` helper was wrong for as long as the same helper signed and verified — which is `ref-21`'s shape as specified.
2. **Carriages.** No rung runs a witnessed exchange with a VSC over v1, v2 and TSP; both landed after every rung this plan cites. Proposed `ref-07i`.
3. **Test surface.** V5 omits `trust-tasks` (`witnessCeremony`, `outcomeEvidence`), the v2/TSP e2e runners, and `ref-15…19`. Also: D4 is trivial to emit but four read sites to fix, three of which (`outcomeEvidence.ts:71`, `:227`, `witnessCeremony.ts:169`) are a WD02 debt we owe regardless — `ref-07e` Act 3 shows they are exactly why 2 of our 4 checks fail on a VWC minted by the Rust catalogue.
4. **Farm.** `feat/prague-farm-membership` forks from `feat/credo-0.7` at `d359e2b`, eleven behind, and is not downstream of this branch; its diff has no witness-shape coupling. But §7.1's `verifiable-trust-infrastructure` fetch repair is also what the local VTI stack needs — one task, two plans waiting on it.

**Two more worth your attention.** `ref-07b` Act 4: `dtg-credentials` 0.7.0 rejects every extension member outright (serde `deny_unknown_fields`) — same class as §7.1's `personhood.rego` but harder, since a compiled-in choice is not a replaceable policy. It doesn't change §3.5, C10 is the normative text, but it's what will actually interoperate at a demo. It also caught a bug of ours that is unrelated to the VSC: we emit `issuer` as `{id, name}` where both drafts require a string. And a small ambiguity in §2.2 — the plan issues under `firstperson.network/credentials/dtg/v1#` while #52's worked entries (what `ref-07f` generates from) use `trustoverip.org/dtg/vocab#`, so "byte-equal to #52's example modulo the namespace constant" needs to say which is which.

**On the conflicts:** not a content disagreement. `feat/credo-0.7` was rebased and force-pushed (all nine commits have new hashes with identical subjects — `17b6330` → `6c2afb9` and so on), so this branch carries pre-rewrite copies of nine commits it doesn't own. A test merge conflicts only in `app/src/hooks/useBCAgentSetup.test.ts` and the `bifold` pointer, both duplication artifacts. Rebasing `acfb85a` + `75eb898` (+ my `39521e0`, all `docs/plans/` only) onto the current tip and force-pushing with lease clears it — happy to run that if you'd rather not.

albertoleon7794:
The rungs `A1` cites are now pushed, so that table is checkable:

- [`feat/ref-07-crossimpl`](https://github.com/berkmancenter/keyring-wallet/tree/feat/ref-07-crossimpl) (`b8dd46c`) — `ref-07b-statement-shape`, `ref-07c-predicate-coverage`, `ref-07d-vocabulary-trust-path`, `ref-07e-crossimpl-vwc`. Branched off `main` at `127fa9c4`, so it is behind current `main`; it also re-signs `ref-07`'s fixtures, which the `eddsa-jcs-2022` fix required.
- [`feat/ref-07f-predicate-registry`](https://github.com/berkmancenter/keyring-wallet/tree/feat/ref-07f-predicate-registry) (`c60fddb`) — `ref-07f`, committed off `main` at `a45364a`, `node run.mjs` green with 0 failures just now. Its ladder row is in `tsp-reference/README.md` in the same commit.

The ones bearing directly on this plan: `ref-07c` Checks 1–4 for §3.7, `ref-07f` Acts 3–4 and 6–7 for §2.2/§2.3 and V3's dispatch, `ref-07b` Act 4 for §3.5's interoperability, and `ref-07e` Acts 3–4 for D3/D4.

Neither branch is a PR yet, and neither is rebased onto current `main` — say the word if you'd rather review them as PRs than as branches.

